PR #12, opened 2024-09-11
feat(mpi): redesign mpi schema

### LINKED ISSUE #8: Models for new schema
## Summary
Create a new `src/record_linker/models.py` module that defines the following schema using the [SQLAlchemy ORM](https://docs.sqlalchemy.org/en/20/orm/quickstart.html).

## Acceptance Criteria
- [ ] README has clear instructions on how to initialize a new database
- [ ] Python classes representing the tables outlined before
- [x] Alembic migrations defined in `migrations/` for initializing the database

### Schema
```mermaid
erDiagram
    Person {
        int id
        uuid internal_id
    }

    ExternalPerson {
        int id
        int person_id 
        string source
        string external_id
    }

    Patient {
        int id
        int person_id
        string data
    }
    
    BlockingKey {
        int id
        int patient_id
        string key
        string value
    }

    Person ||--o{ ExternalPerson: "has"
    Person ||--o{ Patient : "has"
    Patient ||--o{ BlockingKey : "has"
```
We'll also need a new way to manage migrations for the new schema, as [pyway doesn't support Microsoft SQL Server](https://pypi.org/project/pyway/).  For this, we should use [Alembic](https://alembic.sqlalchemy.org/en/latest/) since it has strong support for SQLAlchemy and all the databases it supports.

Create a new [migration environment](https://alembic.sqlalchemy.org/en/latest/tutorial.html) (we should be able to continue to use the `/migrations` directory, as pyway and alembic will read that differently) for the project, and generate a migration for the new ORM created.  The migrations need to be able to use the existing `MPI_*` environment variables that are defined in `.env` for knowing which database to run the migrations on.

## Dependencies
- https://pypi.org/project/alembic/

alhayward:
## Description
* Redefine the Master Patient Index (MPI) database schema using [SQLAlchemy ORM](https://docs.sqlalchemy.org/en/20/orm/quickstart.html)
* Create an [Alembic](https://alembic.sqlalchemy.org/en/latest/tutorial.html) migration environment
* Generate an initial migration for the new schema

### Schema
```mermaid
erDiagram
    Person {
        int id
        uuid internal_id
    }

    ExternalPerson {
        int id
        int person_id 
        string source
        string external_id
    }

    Patient {
        int id
        int person_id
        string data
    }
    
    BlockingKey {
        int id
        string key
    }

    BlockingValue {
        int id
        int patient_id
        int blockingkey_id
        string value
    }

    Person ||--o{ ExternalPerson: "has"
    Person ||--o{ Patient : "has"
    Patient ||--o{ BlockingValue : "has"
    BlockingKey ||--o{ BlockingValue: "has"
```

## Related Issues
Closes #8

## Checklist

- [X] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [X] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [X] I have updated the documentation, if applicable.
- [X] I have added or updated test cases to cover my changes, if applicable.
- [X] I have minimized the number of reviewers to include only those essential for the review.
- [X] I have notified teammates in the review thread to build awareness.

## Checklist for Reviewers


src/recordlinker/models.py
```diff
@@ -0,0 +1,41 @@
...
+class Person(Base):
+    __tablename__ = "mpi_person"
+
+    id: orm.Mapped[int] = orm.mapped_column(primary_key=True)
+    internal_id: orm.Mapped[uuid.UUID] = orm.mapped_column(default=uuid.uuid4)
+
+
+class ExternalPerson(Base):
+    __tablename__ = "mpi_external_person"
+    # Create a unique index on the combination of person_id and external_id
```
alhayward, line 19: Why do we want to do this?

src/recordlinker/models.py
```diff
@@ -0,0 +1,41 @@
...
+    person_id: orm.Mapped[int] = orm.mapped_column(ForeignKey("mpi_person.id"))
+    data: orm.Mapped[dict] = orm.mapped_column(JSON)
+
+
+class BlockingKey(Base):
+    __tablename__ = "mpi_blocking_key"
+
+    id: orm.Mapped[int] = orm.mapped_column(primary_key=True)
+    patient_id: orm.Mapped[int] = orm.mapped_column(ForeignKey("mpi_patient.id"))
+    key: orm.Mapped[str] = orm.mapped_column(String(50), index=True)
```
alhayward, line 40: Why `String(50)` rather than `String(255)` above?

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
ericbuckley, line 1: I think it be great to add some information to this document that outlines the normal procedure for making database schema changes.  Update models.py, run the autogenerate command, run the upgrade command, ...

src/recordlinker/models.py
```diff
@@ -0,0 +1,40 @@
+import uuid
```
ericbuckley, line 1: Small nitpick on the location of this module.

The DIBBs team originally create two packages for this building block container; `app` and `app.linkage`.  From what I gather, the intent was to put the code specific to the algorithm itself and how it functions inside `app.linkage`, where `app` was intended to store code related to the API and processing the data received and sent.  Now there is one glaring omission to this strategy in that all the FHIR parsing code was also stuffed into the `app.linkage` package, as the algorithm really doesn't **need** FHIR to function, that aside though they did a decent job organizing the functionality.

I'd argue that `models.py` is key to how the algorithm works, as it describes the schema necessary for the blocking queries to work.  So for this I'd recommend we move this file to `recordlinker/linkage/models.py`.

src/recordlinker/models.py
ericbuckley: This table will get rather large having the potential for 7x rows for each Patient.  I think we should limit the size of these columns to what we actually need and nothing more.

Since we've started the conversation about limiting the number of blocking keys to a set number (maybe 10 or so), it might make more sense to break out a lookup table here.  A table to store the 10 or so blocking keys that we allow, then have another table like BlockingValue that stores the actually indexes for Patients.

alhayward: I think that makes sense. To make sure I'm understanding, what would be the advantages of creating a separate BlockingValue lookup table (e.g., performance-wise, readability, etc.)?

.env
```diff
@@ -4,3 +4,4 @@ MPI_HOST=localhost
 MPI_PORT=5432
 MPI_USER=postgres
 MPI_PASSWORD=pw
+DB_URI="postgresql+psycopg2://postgres:pw@localhost:5432/testdb"
```
alhayward, line 7: @ericbuckley Should we make a comment/disclaimer here for future contributors?

ericbuckley (approved):
This looks great @alhayward!

ericbuckley:
@alhayward one last thing, please add yourself to the list of contributors in the pyproject.toml file

pyproject.toml
```diff
@@ -11,6 +11,7 @@ authors = [
     {name = "Dan Paseltiner", email = "dan@skylight.digital"},
     {name = "Marcelle Goggins", email = "marcelle@skylight.digital"},
     {name = "Eric Buckley", email = "ericbuckley@skylight.digital"}
```
ericbuckley, line 13:
```suggestion
    {name = "Eric Buckley", email = "ericbuckley@skylight.digital"},
```

ericbuckley (approved):

Issue #10, opened 2024-09-09
Implement MPI client functions for new schema


ericbuckley:
## Summary
Add the `get_block_data` and `insert_matched_patient` functions to a new module, `data_access.py` for the new schema.

## Acceptance Criteria
- [ ] Implementations of the above functions
- [ ] Unit tests for the above functions to show it works as expected

## Details / Tasks
No need to use the existing function signatures for the reimplementation, as there are some readability issues with passing nested lists; going forward, use these signatures

```python
def get_block_data(
    session: orm.Session, 
    data: typing.Dict,
    algo_config: typing.List[typing.Dict],
) -> typing.List[Patient]:
    pass

def insert_matched_patient(
    session: orm.Session,
    data: typing.Dict,
    algo_config: typing.List[typing.Dict],
    person_id: typing.Optional[int],
    external_person_id: typing.Optional[str],
    commit: bool = True,
) -> Patient:
    pass
```

## Notes / Comments

When inserting a patient, abstract the functions necessary for creating the patient row along with creating new blocking keys.  Its likely with this design that we'll need a way to **generate new blocking keys** when an algorithm changes, so keeping that functionality separate will open doors for future expansion.

PR #20, opened 2024-09-16
seperating matching functions into its own module

### LINKED ISSUE #9: Separate transformer and matching functions
## Summary
Create new modules, `linkage/transformers.py` and `linkage/matchers.py`, to store the functions referenced in the [algorithms.py](https://github.com/CDCgov/IDWA-RL/blob/main/src/recordlinker/linkage/algorithms.py) module.

## Acceptance Criteria
- [ ] A transformers.py module with the last4 and first4 functions defined
- [ ] A matchers.py module with all eval_* and feature_* functions defined
- [ ] The existing link.py and mpi.py module should be updated to use these new modules to import functions

## Background / Context

This is not just about making the code reusable to test out the new schema, but also making it easier to read, and setting ourselves up for more flexibility in the future.  When reading the algorithms.py module, its not obvious which are strings and what are references to function names.  Furthermore, the strings that are references to functions, its not clear as to where those functions are defined.  Moving the code to two modules, based on its use cases, will be the first step in clearing up some ambiguity around how its being used in the algorithm configuration.

ericbuckley:
## Description
Moving matching functions into its own module so they can be more easily reused.  Keeping these functions separate, from the logic used for linkage, allows us to move forward with creating an alternative set of linkage functions using a new schema.  Also, changing the syntax of the algorithm configuration for all value references that map to python functions.  The new format is `func:<module_path>.<function_name>`, this not only makes it easy for parsing the algorithm, but also makes it transparent to the author as to where the matching code lives.

## Related Issues
closes #9

## Checklist

## Checklist for Reviewers


src/recordlinker/linkage/matchers.py
```diff
@@ -0,0 +1,427 @@
...
+    :param feature_col: The name of the column being evaluated (e.g. "city").
+    :param col_to_idx: A dictionary mapping column names to the numeric index
+      in which they occur in order in the data.
+    :return: A boolean indicating whether the features are an exact match.
+    """
+    idx = col_to_idx[feature_col]
+    return record_i[idx] == record_j[idx]
+
+
+def feature_match_four_char(
```
alhayward, line 107: What is the purpose of this function when we have transformation functions (e.g., `first4`)?

src/recordlinker/linkage/matchers.py
```diff
@@ -0,0 +1,427 @@
...
+    record_j: list,
+    feature_col: str,
+    col_to_idx: dict[str, int],
+    **kwargs: dict,
+) -> bool:
+    """
+    Determines whether two strings in a given pair of records are close
+    enough to constitute a partial match. The exact nature of the match
+    is determined by the specified string comparison function (see
+    harmonization/utils/compare_strings for more details) as well as a
```
alhayward, line 142: Is this comment out-of-date now that the RL BB is separated from the Harmonization BB?

src/recordlinker/linkage/matchers.py
```diff
@@ -0,0 +1,427 @@
...
+    **kwargs: dict,
+) -> bool:
+    """
+    Determines whether a single feature in a given pair of records
+    constitutes an exact match (perfect equality).
+
+    :param record_i: One of the records in the candidate pair to evaluate.
+    :param record_j: The second record in the candidate pair.
+    :param feature_col: The name of the column being evaluated (e.g. "city").
+    :param col_to_idx: A dictionary mapping column names to the numeric index
```
alhayward, line 99: Out of scope for this PR, but agree w/ past discussions that we can refactor this code to not rely on passing around a `col_to_idx` dict mapping column names to order index. The current implementation is less DRY and readable (which columns are being evaluated?).

src/recordlinker/linkage/matchers.py
```diff
@@ -0,0 +1,427 @@
...
+    """
+    idx = col_to_idx[feature_col]
+
+    # Convert datetime obj to str using helper function
+    if feature_col == "birthdate":
+        record_i[idx] = utils.datetime_to_str(record_i[idx])
+        record_j[idx] = utils.datetime_to_str(record_j[idx])
+
+    # Special case for two empty strings, since we don't want vacuous
+    # equality (or in-) to penalize the score
```
alhayward, line 163: @ericbuckley This comment answers your previous question of: what happens when two fields are missing? How is equality evaluated?

assets/linkage/dibbs_basic_algorithm.json
```diff
@@ -2,8 +2,8 @@
   "algorithm": [
     {
       "funcs": {
-        "first_name": "feature_match_fuzzy_string",
-        "last_name": "feature_match_exact"
+        "first_name": "func:recordlinker.linkage.matchers.feature_match_fuzzy_string",
```
alhayward, line 5: This is out of scope for this PR, but I think renaming the `feature_match_*` and `eval_*` functions down the line would be helpful. The current naming of the functions makes it so it's not entirely clear what they do.

For example, function names that communicate the action/command they are doing (e.g., `fuzzy_match`, `exact_match`) can be clearer I think, rather than those that describe how they are categorized in implementation.

Happy to help out with this in future refactoring efforts!

src/recordlinker/linkage/utils.py
```diff
@@ -73,7 +70,7 @@ def datetime_to_str(
     """
     # Handle None or empty string
     if input_date is None or input_date == "":
-        return input_date
+        return ""
```
alhayward, line 73: What was the reason for this change? (Do we want to make sure two empty strings are evaluated as equal?)

src/recordlinker/linkage/utils.py
```diff
@@ -220,3 +178,45 @@ def get_fhirpathpy_parser(fhirpath_expression: str) -> Callable:
...
+    return bound
+
+
+def str_to_callable(val: str) -> typing.Callable:
+    """
+    Converts a string representation of a function to the function itself.
+    """
+    # Remove the "func:" prefix
+    if val.startswith("func:"):
+        val = val[5:]  # Remove the prefix
```
alhayward, line 209: This inline comment is redundant with line 207 and can probably be removed.

src/recordlinker/linkage/algorithms.py
```diff
@@ -25,30 +25,30 @@
 DIBBS_BASIC = [
```
alhayward, line 25: What is the purpose of having the algorithms defined in multiple places (both `src/recordlinker/linkage/algorithms.py` and `assets/linkage/`)?

alhayward (commented):
Left some comments/questions, some related to future refactoring efforts.

Overall I think this does a really great job of simplifying & making the code more organized!

assets/linkage/dibbs_basic_algorithm.json
ericbuckley: Yeah 100%, I'd also like to change their signatures to accept data objects rather than lists of strings.  Its difficult to understand whats being compared when you have to memorize the order that PII shows up in a list.  Do you want to start a GH issue for this and put it in the backlog?

src/recordlinker/linkage/algorithms.py
ericbuckley: Honestly, not much.  There are some test cases that are using the assets directory for validation, but that could easily be done using `json.dumps` + the contents of algorithms.py.  Long term, many of these test cases will be deleted because the work Cason is doing with moving the algorithms to the DB will make them irrelevant.

src/recordlinker/linkage/matchers.py
ericbuckley: Very similar, but the transformer **alters** data, this function **compares** data.  In issue #10, I'm going to start the process of removing the first4 and last4 transformer functions.  But for now, I'm really just focusing on copying this over from link.py so we can to test the 2 schemas and make sure we're OK to swap our the old one for the new one.  Once we are down to one schema, we'll do some heavy refactoring on this module.

src/recordlinker/linkage/matchers.py
ericbuckley: will change shortly

src/recordlinker/linkage/matchers.py
ericbuckley: Well this is half of it, I'm also interested in how the blocking queries are meant to behave when PII is missing.

src/recordlinker/linkage/utils.py
ericbuckley: There was a potential for a bug here.  The signature of the method states a `str` should be returned, however if the `input_date` is None, `None` could be returned.  I changed it to guarantee that `datetime_to_str` always returns a `str`.

src/recordlinker/linkage/utils.py
ericbuckley: will be updated shortly

tests/unit/test_matchers.py
```diff
@@ -0,0 +1,255 @@
...
+
+    cols = {"first": 0, "last": 1}
+
+    # Simultaneously test matches and non-matches of different data types
+    for c in cols:
+        assert matchers.feature_match_four_char(record_i, record_j, c, cols)
+        assert not matchers.feature_match_four_char(record_i, record_k, c, cols)
+
+
+def test_feature_match_exact():
```
derekadombek, line 161: were these types of cases not being tested before?

alhayward (approved):

PR #31, opened 2024-09-19
Feat/algorithm configuration schema

### LINKED ISSUE #13: New schema for Algorithm configuration data
## Summary
In `models.py` add new classes for storing data related to the available algorithms to run.

## Acceptance Criteria
- [ ] ORM classes to describe the below classes
- [ ] Alembic migrations to create the new tables

## Details / Tasks
```mermaid
erDiagram
    Algorithm {
        int id
        bool is_default "a check should be added to guarentee that only 1 row in the table is marked as the default"
        string label "should be unique"
        string description
    }

    AlgorithmPass {
        int id
        int algorithm_id
        int[] blockingkeys "a list of values from the BlockingKey table"
        string[] evaluators "a list of matching functions and values to use"
        string rule "the evaluation rule function"
        float cluster_ratio
        json kwargs "extra parameters to pass to the evalator functions"
    }

    Algorithm ||--o{ AlgorithmPass: "has"
```

## Dependencies
#8 is adding the Alembic dependency and laying the groundwork for database migrations

cbrinson-rise8:
## Description
In models.py add new classes for storing data related to the available algorithms to run.

```mermaid
erDiagram
    Algorithm {
        int id
        bool is_default "a check should be added to guarentee that only 1 row in the table is marked as the default"
        string label "should be unique"
        string description
    }

    AlgorithmPass {
        int id
        int algorithm_id
        int[] blockingkeys "a list of values from the BlockingKey table"
        string[] evaluators "a list of matching functions and values to use"
        string rule "the evaluation rule function"
        float cluster_ratio
        json kwargs "extra parameters to pass to the evalator functions"
    }

    Algorithm ||--o{ AlgorithmPass: "has"
```

## Related Issues
closes #13

## Additional Notes

## Checklist

## Checklist for Reviewers


.env
```diff
@@ -1 +1,7 @@
-DB_URI="postgresql+psycopg2://postgres:pw@localhost:5432/postgres"
+MPI_DB_TYPE=postgres
```
ericbuckley, line 1: @cbrinson-rise8 these look like the old settings, I think you may have had a merge resolution issue.  This file should just be 1 line.  https://github.com/CDCgov/RecordLinker/blob/main/.env

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +44,48 @@ class BlockingValue(Base):
     patient_id: orm.Mapped[int] = orm.mapped_column(ForeignKey("mpi_patient.id"))
     blockingkey_id: orm.Mapped[int] = orm.mapped_column(ForeignKey("mpi_blocking_key.id"))
     value: orm.Mapped[str] = orm.mapped_column(String(50), index=True)
+
+class Algorithm(Base):
+    __tablename__ = "algorithm"
+
+    id: orm.Mapped[int] = orm.mapped_column(primary_key=True)
+    is_default: orm.Mapped[bool] = orm.mapped_column(default=False)
```
ericbuckley, line 52:
```suggestion
    is_default: orm.Mapped[bool] = orm.mapped_column(default=False, index=True)
```

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +44,48 @@ class BlockingValue(Base):
...
+    label: orm.Mapped[str] = orm.mapped_column(String(255), unique=True)
+    description: orm.Mapped[str]
```
ericbuckley, line 54:
```suggestion
    description: orm.Mapped[str] = orm.mapped_column(Text())
```

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +44,48 @@ class BlockingValue(Base):
...
+    Parameters:
+    connection: The database connection being used for the operation.
+    target: The instance of the Algorithm class being inserted or updated.
+    
+    Raises:
+    Exception: If another algorithm is already marked as default.
+    """
+     
+    if target.is_default:
+        existing_default = connection.execute("SELECT COUNT(*) FROM algorithm WHERE is_default = TRUE").scalar()
```
ericbuckley, line 75: Eventually, we'll be adding an optional configuration settings to the project for database table prefixes via a [declared_attr](https://docs.sqlalchemy.org/en/20/orm/mapping_api.html#sqlalchemy.orm.declared_attr) or something like that.

We know that NBS would like to have these tables live in an existing Microsoft SQL Server, and other STLTs may want to do the same down the road.  This opens up the potential for table name clashes, as the NBS team may already have a table named "algorithm" on their server.

If we stick to using the ORM, we can make this dynamic and allow the users to add/change the default prefix if they wish so, to prevent any table name clashes.

Could you rewrite this query using the ORM?

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +44,48 @@ class BlockingValue(Base):
...
+        if(existing_default > 0):
```
ericbuckley, line 76: This will fail to catch the "before_update" use case.  If the user is updating an Algorithm, that is already set as the default.  If the update say, changes the description on the algorithm, but leave the `is_default=True` as is.  This if block will pass and the update will fail.

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +44,48 @@ class BlockingValue(Base):
...
+            raise ValueError("There can only be one default algorithm.")
+
+event.listen(Algorithm, 'before_insert', check_only_one_default)
```
ericbuckley, line 79: Could we add some test cases for these situations to make sure the `check_only_one_default` works as expected.

alembic/versions/0db31a429322_create_algorithm_tables.py
```diff
@@ -0,0 +1,51 @@
...
+down_revision: Union[str, None] = '6052c193a26a'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('algorithm',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('is_default', sa.Boolean(), nullable=False),
```
alhayward: Should there be a constraint that ensures only 1 row in the `algorithm` table has `is_default` value of `true`?

alembic/versions/0db31a429322_create_algorithm_tables.py
```diff
@@ -0,0 +1,51 @@
...
+    sa.Column('label', sa.String(length=255), nullable=False),
+    sa.Column('description', sa.String(), nullable=False),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('label')
+    )
+    op.create_table('algorithm_pass',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('algorithm_id', sa.Integer(), nullable=False),
+    sa.Column('blocking_keys', sa.JSON(), nullable=False),
+    sa.Column('evaluators', sa.JSON(), nullable=False),
```
alhayward: #13 indicates that the `blocking_keys` column should be list of int and that the `evaluators` column should be list of str. What was the reason for the change to JSON type?

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +45,45 @@ class BlockingValue(Base):
     patient_id: orm.Mapped[int] = orm.mapped_column(ForeignKey("mpi_patient.id"))
     blockingkey_id: orm.Mapped[int] = orm.mapped_column(ForeignKey("mpi_blocking_key.id"))
     value: orm.Mapped[str] = orm.mapped_column(String(50), index=True)
+
+class Algorithm(Base):
+    __tablename__ = "algorithm"
```
alhayward, line 50: Is there a reason why this table name isn't prefixed with `mpi_`? Should we prefix with something else?

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +45,45 @@ class BlockingValue(Base):
...
+
+    id: orm.Mapped[int] = orm.mapped_column(primary_key=True)
+    is_default: orm.Mapped[bool] = orm.mapped_column(default=False, index=True)
+    label: orm.Mapped[str] = orm.mapped_column(String(255), unique=True)
+    description: orm.Mapped[str] = orm.mapped_column(Text())
```
alhayward, line 55: Why `Text()` type rather than `String()`?

src/recordlinker/linkage/models.py
cbrinson-rise8: updated

src/recordlinker/linkage/models.py
cbrinson-rise8: updated

.env
cbrinson-rise8: good catch! Changed it back to be the same as the .env on main

alembic/versions/3a677cb88b01_create_algorithm_tables.py
```diff
@@ -0,0 +1,53 @@
...
+down_revision: Union[str, None] = '6052c193a26a'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('algorithm',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('is_default', sa.Boolean(), nullable=False),
```
alhayward, line 25: Should there be a constraint that ensures only 1 row in the `algorithm` table has `is_default` value of `true`?

alembic/versions/3a677cb88b01_create_algorithm_tables.py
```diff
@@ -0,0 +1,53 @@
...
+    sa.Column('description', sa.Text(), nullable=False),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('label')
+    )
+    op.create_index(op.f('ix_algorithm_is_default'), 'algorithm', ['is_default'], unique=False)
+    op.create_table('algorithm_pass',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('algorithm_id', sa.Integer(), nullable=False),
+    sa.Column('blocking_keys', sa.JSON(), nullable=False),
+    sa.Column('evaluators', sa.JSON(), nullable=False),
```
alhayward, line 36: #13 indicates that the `blocking_keys` column should be list of int and that the `evaluators` column should be list of str. What was the reason for the change to JSON type?

alhayward (commented):
Nice job!

src/recordlinker/linkage/models.py
ericbuckley: String is based on the Varchar type, so a max length is required.  Text is a flexible SQL type that allows unlimited string length.  https://www.sprinkledata.com/blogs/postgresql-text-vs-varchar-choosing-the-right-data-type-for-your-database

alembic/versions/3a677cb88b01_create_algorithm_tables.py
ericbuckley: [Partial indexes are cool](https://www.postgresql.org/docs/current/indexes-partial.html), and hopefully one day is becomes standard ANSI SQL, but to my knowledge Postgresql is the only RDMS we are planning on supporting that has this feature.  Alternatively, Cason added a function in models.py that should prevent users from making this mistake.

alembic/versions/3a677cb88b01_create_algorithm_tables.py
ericbuckley: Arrays/Lists aren't a standard Type in SQL.  This is really by design, since traditional database schemas require new tables for one->many relationships.  So 20 years ago we would have needed to create a `algorithm_pass_blockingkey` and `algorithm_pass_evaluator` child tables to store this.


We could always go that route, but for small datasets, I find it less readable (you need to go back and forth between tables, you always have to join tables when querying, you have to remember to add order columns to preserve order).  Lists kind of make this easier.

While JSON is becoming more commonplace, postgresql, mysql, microsoft sql server, sqlite3 all support this.  ARRAY is not.  So for maximum database support, using JSON gives us more flexibility.

src/recordlinker/linkage/models.py
cbrinson-rise8: I left it off because I was thinking of it as something different than an MPI. I can add it back just for consistency with the other tables unless we have any other ideas for the name?

alembic/versions/3a677cb88b01_create_algorithm_tables.py
cbrinson-rise8: I had it as a constraint initially but it would have needed to be changed for any other type of DB. There's a function that is run whenever an `update` or `insert` happens on the algorithm table that should take care of this.

alembic/versions/3a677cb88b01_create_algorithm_tables.py
cbrinson-rise8: Postgresql does support Array datatypes but I made it JSON for the same reasons Eric said above^

src/recordlinker/linkage/models.py
ericbuckley: I've thought about this as well.  I think longer term we might have a structure like `models/mpi.py`, `models/config.py`, `models/training.py`, etc as we start to build out more features in the product.  At that time, distinct prefixes for each submodule makes sense.  I'm also fine with whats here now, and punting that decision to the future when we have more clarity.

src/recordlinker/linkage/models.py
cbrinson-rise8: @ericbuckley Was looking at adding a test_models.py file as a place to put the test for this function but it looks like you created one in https://github.com/CDCgov/RecordLinker/pull/30

ericbuckley: test_models.py makes sense, we can rebase appropriately after one of us merges.

src/recordlinker/linkage/models.py
```diff
@@ -47,3 +45,46 @@ class BlockingValue(Base):
...
+    target: The instance of the Algorithm class being inserted or updated.
+    
+    Raises:
+    ValueError: If another algorithm is already marked as default.
+    """
+
+    session = orm.Session.object_session(target)
+  
+    if target.is_default:
+        existing = session.query(Algorithm).filter(Algorithm.is_default == True).first()    # noqa
```
cbrinson-rise8, line 73: `# noqa` suppreses the ruff linting error on this line. The reason for this is ruff wants to enforce the use of the `is` instead of the `==` operator. But using `is` doesn't seem to work when comparing SQL boolean values.

cbrinson-rise8:
Just a note for anyone wondering what `# noqa` is. It suppreses the ruff linting error on this line. The reason for this is ruff wants to enforce the use of the `is` instead of the `==` operator. But using `is` doesn't seem to work when comparing SQL boolean values.

src/recordlinker/linkage/models.py
ericbuckley, line 73: @cbrinson-rise8 your comment on why this should be ignored is really helpful, mind putting that in the code for future eyes?  Also if we can just ignore that 1 rule, that be great.
```suggestion
        existing = session.query(Algorithm).filter(Algorithm.is_default == True).first()    # noqa: E712
```

cbrinson-rise8: added comment explanation and the specific rule number

ericbuckley (approved):
Looks great, thanks for all your work on this @cbrinson-rise8

ericbuckley:
@cbrinson-rise8 one more small thing.  Can you please add your name to the [authors list](https://github.com/CDCgov/RecordLinker/blob/main/pyproject.toml#L9)!

PR #34, opened 2024-09-23
fix(match): normalize DOB type

### LINKED ISSUE #17: Duplicate patients with multiple last names not matching
## Summary
When sending through the attached patient payload twice to the link-record endpoint, both calls result in a "match not found" result.  The algorithm should find the second patient record, which has the exact same PII as the first, matches the person generated in the first call.

## Impact
This could be a bug in the algorithm, but it could also be a design flaw of DIBBS_BASIC.  We need to investigate further to see what exactly is causing this to be not a match.

## Steps to reproduce
1. Start up the API server using scripts/local_server.sh
2. Download the attached Patient bundle json file
3. Send the payload to the API server with curl
    a. curl -X POST --header "Content-Type: application/json" -d @bundle.json "http://localhost:8000/link-record"
4. The response should indicate no match was found (this is correct since the database was empty)
5. Edit the bundle.json file to have a new UUID on line 9
6. Repeat step 3
7. The response should indicate no match was found (this is **incorrect** since we should match the existing patient in the database)

## Expected behavior
No match on the first call and a **match** on the second call

## Additional context
[bundle.json](https://github.com/user-attachments/files/16998396/bundle.json)

alhayward:
## Description
* Normalize `birthdate` type of an existing Patient record from `datetime.time` to `str` upon match evaluation, before comparing with `birthdate` of an incoming Patient record.

## Related Issues
* Closes #17 

## Additional Notes
Initially, two of the unit tests failed as a result of the new logic after resolving the bug. By logging the similarity scores & match decisions, I understood why the tests were wrongfully succeeding previously. To address this, I changed the expected results of these two tests, rather than change the FHIR bundle of test cases itself, to avoid downstream test failure where this bundle is used elsewhere.

In the future, we should normalize `birthdate` type elsewhere, and only once. Currently, this `datetime.time` to `str` type conversion happens several places. This can be done more DRYly, and any data transformations/normalizations of blocking or matching fields should be done separately from where blocking or matching occurs, respectively.

### DIBBs Basic
1. Start up the API server by running the following command in the root directory in the CLI:
```
scripts/local_server.sh
```
2. Download the attached Patient record ([bundle.json](https://github.com/user-attachments/files/16998396/bundle.json) file)
3. Send the payload to the API server by running the following command:
```
curl -X POST --header "Content-Type: application/json" -d @bundle.json "http://localhost:8000/link-record"
```
The response should indicate no match was found (this is correct, since the database was empty)

4. Edit the [bundle.json](https://github.com/user-attachments/files/16998396/bundle.json) file to have a new UUID on line 9
5. Repeat Step 3
The response should indicate a match _was_ found (this is correct, since we should match the existing patient in the database)

### DIBBs Enhanced
1. To test the DIBBs Enhanced algorithm, edit the [bundle.json](https://github.com/user-attachments/files/16998396/bundle.json) file to add a second top-level key of `"use_enhanced: true"` (on the same level as the `"bundle"` key)
2. Repeat Steps 1, 3, 4, and 5 above

## Checklist

## Checklist for Reviewers

### DIFF
--- a/tests/unit/test_linkage.py
+++ b/tests/unit/test_linkage.py
@@ -474,14 +474,14 @@ def test_link_record_against_mpi():
     # First patient inserted into empty MPI, no match
     # Second patient blocks with first patient in first pass, then fuzzy matches name
     # Third patient is entirely new individual, no match
-    # Fourth patient fails blocking with first pass then fails on second
-    # Fifth patient: in first pass MRN blocks with one cluster but fails name,
-    # in second pass name blocks with different cluster but fails address, no match
-    # Sixth patient: in first pass, MRN blocks with one cluster and name matches in it,
-    # in second pass name blocks on different cluster and address matches it,
-    # finds greatest strength match and correctly assigns to larger cluster
-    assert matches == [False, True, False, False, False, False]
-    assert sorted(list(mapped_patients.values())) == [1, 1, 1, 1, 2]
+    # Fourth patient fails blocking in first pass, blocks with first patient in second
+    # pass, then fuzzy matches on address and exact matches on birthdate, joins cluster
+    # with first and second patient
+    # Fifth patient: fails blocking in first and second pass, no match
+    # Sixth patient: fails blocking in first pass, blocks with fifth patient in second pass,
+    # then matches on birthdate but fails on address, no match
+    assert matches == [False, True, False, True, False, False]
+    assert sorted(list(mapped_patients.values())) == [1, 1, 1, 3]
 
     # Re-open connection to check for all insertions
     patient_records = MPI.dal.select_results(select(MPI.dal.PATIENT_TABLE))
--- a/src/recordlinker/linkage/matchers.py
+++ b/src/recordlinker/linkage/matchers.py
@@ -101,6 +101,12 @@ def feature_match_exact(
     :return: A boolean indicating whether the features are an exact match.
     """
     idx = col_to_idx[feature_col]
+
+    # Convert datetime obj to str using helper function
+    if feature_col == "birthdate":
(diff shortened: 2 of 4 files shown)


ericbuckley:
Thanks for adding this @alhayward, I think the new solution should address this as well because we'll be guaranteeing that all input for comparisons are strings.  Would you also be able to add a test case that highlights this, just to make sure that the new solution has coverage as well.

tests/unit/test_linkage.py
```diff
@@ -474,14 +474,14 @@ def test_link_record_against_mpi():
...
-    assert matches == [False, True, False, False, False, False]
-    assert sorted(list(mapped_patients.values())) == [1, 1, 1, 1, 2]
+    # Fourth patient fails blocking in first pass, blocks with first patient in second
+    # pass, then fuzzy matches on address and exact matches on birthdate, joins cluster
+    # with first and second patient
+    # Fifth patient: fails blocking in first and second pass, no match
+    # Sixth patient: fails blocking in first pass, blocks with fifth patient in second pass,
+    # then matches on birthdate but fails on address, no match
+    assert matches == [False, True, False, True, False, False]
+    assert sorted(list(mapped_patients.values())) == [1, 1, 1, 3]
```
ericbuckley, line 484: @alhayward thanks for the fix here, I just looked at the two records in question ([pat1](https://github.com/CDCgov/RecordLinker/blob/main/assets/patient_bundle_to_link_with_mpi.json#L9-L47) and [pat4](https://github.com/CDCgov/RecordLinker/blob/main/assets/patient_bundle_to_link_with_mpi.json#L153-L191)), and yeah clearly thats a match on the second linkage pass.  Good work.

ericbuckley (approved):

Issue #15, opened 2024-09-12
Update API with Algorithm configuration changes


ericbuckley:
## Summary
Update the API link-record endpoint to only accept one argument for specifying the algorithm, and create a new endpoint to list the available algorithms.

## Acceptance Criteria
- [x] create a new endpoint, `algorithms` that will list the available algorithms based on the database configuration
- [x] update the `link-record` endpoint to accept just 1 parameter for the algorithm to use
- [x] test cases for the changes above
- [x] update OpenAPI docs for the endpoint changes

## Details / Tasks
- Remove the use_enhanced and algo_config parameters on the `link-record` endpoint and replace that with a new parameter, `algorithm`.  The value should be a string, that maps to the unique label stored in the Algorithm table (ie DIBBS_BASIC or DIBBS_ENHANCED).
- When the `link-record` receives a request with no algorithm specified, the default should be used.
- When an invalid algorithm is specified (meaning no matching label in the DB), a 422 should be return with an explanation
- For valid requests, the Algorithm should be **sourced from the algorithms.py** file.  This is a short-term change, as we need to update the code in link.py and mpi.py for parsing Algorithm data before we can pass through an instance of the Algorithm model

## Dependencies
#13 we'll need to be able to query the new Algorithm database table first.

## Testing Considerations
Consider using the [built-in FastAPI test client](https://fastapi.tiangolo.com/reference/testclient/) for stubbing out new test cases for the API changes.  Don't worry about integration tests at this time, unit tests are the priority.  We can mock out any calls to the `link_record_against_mpi` and `add_person_resource` functions.

PR #54, opened 2024-10-01
Feat/algo config api update

cbrinson-rise8:
## Description
Update the API link-record endpoint to only accept one argument for specifying the algorithm, and create a new endpoint to list the available algorithms.

- [x] create a new endpoint, `algorithms` that will list the available algorithms based on the database configuration
- [x] update the `link-record` endpoint to accept just 1 parameter for the algorithm to use
- [x] test cases for the changes above
- [x] update OpenAPI docs for the endpoint changes

## Related Issues
closes. #15 

## Checklist

## Checklist for Reviewers


ericbuckley:
@cbrinson-rise8 make sure to add your name to the [authors list](https://github.com/CDCgov/RecordLinker/blob/main/pyproject.toml#L9-L15) in this PR!

src/recordlinker/linking/mpi_service.py
```diff
@@ -107,3 +107,25 @@ def insert_blocking_keys(
     if commit:
         session.commit()
     return values
+
+def get_all_algorithm_labels(session: orm.Session) -> list[str]:
```
ericbuckley, line 111: I'm thinking we should have a separate service module for these functions.  The intention of this module was just to interact with the `mpi_*` tables.

src/recordlinker/linking/mpi_service.py
```diff
@@ -107,3 +107,25 @@ def insert_blocking_keys(
...
+    """
+    Gets a list of algorithms from the MPI database.
+    returns: list of all labels column from algorithms table
+    """
+    algorithms_list = session.query(models.Algorithm.label).all()
```
ericbuckley, line 116: [scalars](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.scalars) can be a shortcut here if you want, nothing wrong with this implementation, but just so you are aware.

```python
return session.scalars(session.query(models.Algorithm.label)).all()
```

src/recordlinker/linking/mpi_service.py
```diff
@@ -107,3 +107,25 @@ def insert_blocking_keys(
...
+
+    return [algorithm[0] for algorithm in algorithms_list]
+
+def get_algorithm_by_label(session: orm.Session, label: str) -> models.Algorithm | None:
```
ericbuckley, line 120:
```suggestion
def get_algorithm_by_label(session: orm.Session, label: str | None) -> models.Algorithm | None:
```

src/recordlinker/linking/mpi_service.py
```diff
@@ -107,3 +107,25 @@ def insert_blocking_keys(
...
+    """
+    Gets a single algorithm by searching for the unique label
+    returns: algorithm object
+    """
+    if not label:
+        algorithm = session.query(models.Algorithm).filter(models.Algorithm.is_default == True).first() # noqa: E712
```
ericbuckley, line 126: I think this should be a `get()`, instead of a `filter()`.  There shouldn't be a situation where there is not one, and only one row, with is_default == True.  If it's missing or if there are more than one row with is_default set to true, I think we should be raising an error.

src/recordlinker/main.py
```diff
@@ -99,6 +89,14 @@ class HealthCheckResponse(BaseModel):
         description="Returns status of connection to Master Patient Index(MPI)"
     )
 
+class GetAlgorithmsResponse(BaseModel):
+    """
+    The schema for response fromt he record linkage get algorithms endpoint
```
ericbuckley, line 94:
```suggestion
    The schema for response from  he record linkage get algorithms endpoint
```

src/recordlinker/main.py
```diff
@@ -201,3 +210,13 @@ async def link_record(
             "updated_bundle": input_bundle,
             "message": f"Could not connect to database: {err}",
         }
+
+@app.get("/algorithms")
+async def get_algorithms() -> GetAlgorithmsResponse:
```
ericbuckley, line 215: What do you think about renaming this to be get_algorithm_labels?  Technically we're not returning the whole algorithm, just the available labels.

src/recordlinker/linking/mpi_service.py
cbrinson-rise8: Good callout. I like this way much better. Also, I got a deprecation warning when running tests so I went ahead and switched it to use `select` instead of `session.query`. Mainly just to get rid of the warning

ericbuckley: Yeah, I've seen a few of those warnings as well in the past.  The 2.0 style is a bit different, but we should probably start moving in that direction https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-orm-usage

src/recordlinker/linking/mpi_service.py
cbrinson-rise8: I went ahead and switched these to `select` as well with `session.scalar`. You can do a `session.get(...)` but I think you can only use the primary key in that case.

src/recordlinker/linking/mpi_service.py
cbrinson-rise8: I can move these to a new file. Any ideas for the name? Something like `algorithm_service.py`, `config_service.py`, etc?

src/recordlinker/linking/mpi_service.py
ericbuckley: You're right, [get()](https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.get) only works with primary keys, we'll have to build in this logic ourselves or put a helper function on the Algorithm class.

src/recordlinker/linking/mpi_service.py
ericbuckley: algorithm_service.py works for me

tests/unit/linking/test_algorithm_service.py
```diff
@@ -0,0 +1,68 @@
...
+def test_get_all_algorithm_labels(session):
+    testLabel = "DIBBS_BASIC"
+    algo1 = models.Algorithm(label=testLabel, is_default=True, description="First algorithm")
+    session.add(algo1)
+    session.commit()
+    
+    algorithmsList = algorithm_service.get_all_algorithm_labels(session)
+    assert algorithmsList == [testLabel]
+
+class TestGetAlgorithmByLabel:
```
ericbuckley, line 40: Thanks for writing this test suite!

ericbuckley (approved):
Looks good, thanks for all the unit tests!

PR #66, opened 2024-10-03
updating link-record endpoint to use either schema

### LINKED ISSUE #24: Parameterize main for different schemas
## Summary
Add a parameter to the link-record API (or an environment variable), so we can easily test out the performance differences between link.py and simple_link.py

## Acceptance Criteria
- [ ] A parameter or env variable to swap between the two modules
- [ ] Unit test cases for the above switch

## Dependencies
#23

ericbuckley:
## Description
Making changes to main.py link-record route so the user can opt to use either of the two schemas, and the code required to interact with them, via a request header.  Also adding OpenTelemetry so we can trace the linking algorithm and its performance.

## Related Issues
closes #24 

## Additional Notes
* The X-Use-Simple-Link request header can be sent with a true value to indicate the new schema should be used.
* added `src/recordlinker/database.py` and moved session creation code there.
* updated the get_block_data function to use the new sqlalchemey querying syntax and resolved an issue where the query could return duplicate rows

## Checklist

## Checklist for Reviewers


src/recordlinker/linking/link.py
```diff
@@ -139,43 +140,46 @@ def link_record_against_mpi(
...
+            with TRACER.start_as_current_span("link.evaluate"):
+                for person, patients in clusters.items():
+                    matched_count = 0
+                    for patient in patients:
+                        # increment our match count if the pii_record matches the patient
+                        with TRACER.start_as_current_span("link.compare"):
+                            if compare(pii_record, patient, linkage_pass):
+                                matched_count += 1
+                    # calculate the match ratio for this person cluster
+                    match_ratio = matched_count / len(patients)
```
cbrinson-rise8, line 165: I don't think it would ever happen but could `len(patients)` ever be 0? I'm thinking no since you need at least one patient for a cluster. But just thought of that edge case since its dividing by length here

ericbuckley: It's a good thought.  It's not possible now, at least from what I can read of the code.  If you look at this [block where the clusters are created](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/linking/link.py#L145-L152), a Person can only be created if a Patient record already exists.  Therefore, at a minimum, the length has to be 1.

However, there is a chance in the future that someone changes this block, which introduces a case where it could be of length 0, in which case we'd get this weird ZeroDivisionError.  Probably safest to put an assertion at the top of this loop to make sure the length isn't 0.  A custom error message will help debug that if the clustering logic ever gets changed.

ericbuckley: added. https://github.com/CDCgov/RecordLinker/pull/66/commits/2ee6d5ec9feff4b78e31bfbd0fe845efc4953767

src/recordlinker/linking/mpi_service.py
```diff
@@ -21,7 +21,7 @@ def get_block_data(
     blocking keys defined in the algo_config.
     """
     # Create the base query
-    query = session.query(models.Patient)
+    expr = expression.Select(models.Patient.id).distinct()
```
cbrinson-rise8, line 24: Could probably import `select` straight from `sqlalchemy`. (Only mentioning this because I used the `select` somewhere else in the `algorithm_service.py` code). I don't think it matters functionally though.

ericbuckley: They're both defined in [sqlalchemy.sql.expressions](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/expression.py), so availability of them isn't an issue.

Looks like the [select() function just invokes Select.__init__() class](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/_selectable_constructors.py#L451-L494).  TBH, I'm not sure what the point is, I think maybe for some backwards compatibility between the 1.x and 2.x styles?

I'm going to make the change though, since the [style guide](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-orm-usage) does show examples of the function, but not the class.

ericbuckley: changed. https://github.com/CDCgov/RecordLinker/pull/66/commits/88a28d5b502171626352501771841637077edc8c

cbrinson-rise8: Okay sounds good! Yeah I'm hoenstly not sure what the difference is either besides I see `select()` in the current documentation

cbrinson-rise8 (approved):
Looks good!

PR #72, opened 2024-10-09
Update to use algo object

### LINKED ISSUE #16: Update link.py and mpi.py to use Algorithm object
## Summary
The link.py and mpi.py modules expect a Dict for the algo_config parameter, update these to be an instance of the Algorithm class in models.py.  We may also need to rewrite how the simple_mpi.py and simple_link.py use the algo_config, but only if those modules have been committed to main by the time this issue is worked on.

## Acceptance Criteria
- [ ] Update all references to the algo_config, and its parts, in link.py to use Algorithm
- [ ] Update all references in mpi.py to using parts of an algo_config to using either an Algorithm or AlgorithmPass
- [ ] update any test cases that specify the use of an algo_config Dict to an instance of Algorithm
- [ ] Update the API in main.py to pass in an instance of Algorithm to the `link_record_against_mpi` function

## Dependencies
#15

cbrinson-rise8:
## Description
The link.py and mpi.py modules expect a Dict for the algo_config parameter, update these to be an instance of the Algorithm class in models.py. We may also need to rewrite how the simple_mpi.py and simple_link.py use the algo_config, but only if those modules have been committed to main by the time this issue is worked on.

## Related Issues
closes #16 

## Additional Notes
- Removes uses of the `algo_config` `dict` in the `src.recordlinker.linking` directory and replaces them with uses of `Algorithm` and/or `AlgorithmPass` objects from our database schema

## Checklist

## Checklist for Reviewers


src/recordlinker/linking/link.py
```diff
@@ -82,16 +82,16 @@ def add_person_resource(
...
 
 
-def compare(record: models.PIIRecord, patient: models.Patient, linkage_pass: dict) -> bool:
+def compare(record: models.PIIRecord, patient: models.Patient, algorithm_pass: models.AlgorithmPass) -> bool:
     """
     Compare the incoming record to the linked patient
     """
     # all the functions used for comparison
-    funcs: dict[models.Feature, matchers.FEATURE_COMPARE_FUNC] = linkage_pass["funcs"]
+    funcs: dict[models.Feature, matchers.FEATURE_COMPARE_FUNC] = utils.bind_functions(algorithm_pass.evaluators)
```
ericbuckley, line 90: I was thinking we'd do this a bit earlier.  I think likely when we pull it out of the database, but I haven't really thought that through, so maybe just doing it at the top of link is safest for now.  Anyway, if we have 20 patient records come back, 2 passes and 2 evaluation features to compare, we call this method 80 times.  If we do the binding earlier in the stack, we can reduce the bind_functions overhead.

src/recordlinker/linking/link.py
```diff
@@ -82,16 +82,16 @@ def add_person_resource(
...
     # a function to determine a match based on the comparison results
-    matching_rule: matchers.MATCH_RULE_FUNC = linkage_pass["matching_rule"]
-    # keyword arguments to pass to comparison functions and matching rule
-    kwargs: dict[typing.Any, typing.Any] = linkage_pass.get("kwargs", {})
+    matching_rule: matchers.MATCH_RULE_FUNC = utils.str_to_callable(algorithm_pass.rule)
```
ericbuckley, line 92: ditto to my above comment on the evaluators

src/recordlinker/linking/mpi_service.py
```diff
@@ -27,12 +27,10 @@ def get_block_data(
...
     # is considered a match.
-    for idx, block in enumerate(algo_config["blocks"]):
-        key_name = block["value"].upper()
-        if not hasattr(models.BlockingKey, key_name):
-            raise ValueError(f"Invalid Blocking Key: {block}")
-        # Get the matching Blocking Key based on the value in the algo_config
-        key = models.BlockingKey[key_name]
+    for idx, key_id in enumerate(algorithm_pass.blocking_keys):
+        #get the BlockingKey obj from the id
+        key = models.BlockingKey.from_id(key_id)
```
ericbuckley, line 32: I think the `if not hasattr(models.BlockingKey, key_name)` check is still a good one here.  Because the data is coming from the model, we should be verifying that the data is valid before persisting now, so at this point in time it really can't be a user error.  But its possible that data is corrupted or some bug is introduced, so an assertion would be helpful here just in case.
```suggestion
        assert hasattr(models.BlockingKey, key_id), "Invalid Blocking Key: {key_id}")
        key = models.BlockingKey.from_id(key_id)
```

src/recordlinker/models/algorithm.py
```diff
@@ -54,7 +54,7 @@ class AlgorithmPass(Base):
     algorithm_id: orm.Mapped[int] = orm.mapped_column(schema.ForeignKey("algorithm.id"))
     algorithm: orm.Mapped["Algorithm"] = orm.relationship(back_populates="passes")
     blocking_keys: orm.Mapped[list[int]] = orm.mapped_column(sqltypes.JSON)
-    evaluators: orm.Mapped[list[str]] = orm.mapped_column(sqltypes.JSON)
+    evaluators: orm.Mapped[dict] = orm.mapped_column(sqltypes.JSON)
```
ericbuckley, line 57: why the change from `list[str]` to `dict`?

tests/unit/conftest.py
```diff
@@ -26,12 +27,120 @@ def client():
...
 
-
 @pytest.fixture
 def basic_algorithm():
-    return utils.read_json_from_assets("linking", "basic_algorithm.json")["algorithm"]
-
+    basic_algo_pass1 = models.AlgorithmPass(
+        id=1, 
+        algorithm_id=1, 
+        blocking_keys=[1,2,3],
```
ericbuckley, line 35: I was thinking we'd store the actual names here of the Blocking Keys for readability sake.  It adds to the storage, but this table is going to be small.  For example

```
blocking_keys=["BIRTHDATE", "SEX", "FIRST_NAME"]
```

src/recordlinker/models/algorithm.py
cbrinson-rise8: Becuase it should have been a `dict` from the start. `evaluators` in the old `DIBBS_BASIC` and `DIBBS_ADVANCED` `algo_config`s where just `dict`.

src/recordlinker/linking/mpi_service.py
cbrinson-rise8: I added a `ValueError` in the `models.BlockingKey.from_id(...)` function that raises if there's no key with the given `key_id`. I was considering adding this assert but I don't think it will recognize the `key_id` as a attribute since its just an `int` and not an actual `BlockingKey` object

tests/unit/conftest.py
cbrinson-rise8: I could make that change. I would just need to change the `AlgorithmPass` model to store a `list[str]` for `blocking_keys`

src/recordlinker/linking/mpi_service.py
cbrinson-rise8: I can go ahead and change `AlgorithmPass` model to store a `list[str]` like you commented about below, then I could remove the `from_id(...)` function and just do the assert statement you suggested above

src/recordlinker/linking/link.py
cbrinson-rise8: I was thinking about doing this but the `utils.bind_functions` takes in a `dict` and modifies the function `str` to a `callable` I put it here to avoid using a dict and keep `AlgorithmPass` everywhere else.

src/recordlinker/models/algorithm.py
ericbuckley: Ah thats right, thanks for the reminder I was confusing this with blocking_keys for a min.  Would you mind making this small change though?
```suggestion
    evaluators: orm.Mapped[dict[str,str]] = orm.mapped_column(sqltypes.JSON)
```

src/recordlinker/linking/link.py
ericbuckley: Ah, that's a great point, if we modify it it's really no longer a AlgorithmPass object to pass around.  I think there is room here for a future optimization with the Algorithm schema class I'm working on, but no need to hold this up for that.  Would you mind just adding a TODO comment here explaining the opportunity for an optimization if we can bind earlier in the stack, then I think we'll be good.

ericbuckley (approved):
Looks good @cbrinson-rise8 thanks for addressing some old mypy issues in main.py

PR #82, opened 2024-10-17
interfaces for configuring algorithms

### LINKED ISSUE #65: Script to load algorithm.yml
## Summary
Create a script that will load a given algorithm.toml file into the database.  If the label already exists, it should replace the existing algorithm.  If it doesn't yet exist, a new algorithm should be created.

## Acceptance Criteria
- [ ] A new script, "algorithm_configure.sh", to manage algorithm configuration from the command line
- [ ] Documentation in `docs/` to describe how to use this script.

## Details / Tasks
This is really a temporary script, long-term we will be developing a UI over the web application that allows users to configure their algorithm(s) in the browser.  However, we do need a short-term solution for users to edit their algorithms.

Create a new script, in the `scripts/` directory, that will read in a [YAML file](https://yaml.org/).  The structure of this file will be very similar to the existing algorithm.json files, however this new format will allow us to put comments in the file to instruct users how to use it.

The script should load in whatever algorithms are defined in the file, into the algorithm* database tables.  The following rules should be applied:
- If the label already exists, we should ask the user if they want to replace the algorithm or exit
- If the label doesn't exist, we should just create a new algorithm
- If the is_default algorithm is set, and an existing algorithm already has it set, we should ask the user if they want to change the default or exit
- If more than one algorithm in the file has is_default set, we should raise an error
- Blocking Keys should be defined by their name (eg BIRTHDATE, MRN, etc), and validated
- Evaluators should be defined by the matcher function path, and validated (only FEATURE_COMPARE functions in the matchers module are allowed)
- The matching rule should be defined by the matcher evaluator function path, and validated (only MATCH_RULE functions in the matchers module are allowed)
- The cluster ratio should be validated as a numeric
- If the argument "CLEAR" is sent to the script, a prompt should ask them if they are sure they want to delete all algorithms before proceeding
- If the arguments "DELETE X", where X is a valid algorithm label, is sent to the script, a prompt should ask them if they are sure they want to delete Algorithm X before proceeding

## Background / Context
This is a good amount of work, for a temporary script.  However, a lot of these validations will eventually be needed for the future Algorithm UI in the future.  If we can structure these validations in a reusable way, say with some pydantic model(s), we can make progress towards that future goal.

## Dependencies
#57

ericbuckley:
## Description
Basic CRUD endpoints for managing algorithms and a new script, scripts/algorithm_config.py, for performing similar tasks on the command line.

## Related Issues
closes #65 

## Additional Notes
- `schemas.algorithm` was created for parsing, validating and serializing Algorithm data.  The addition of `AlgorithmSummary` was added specifically for the listing view API so we can summarize what the Algorithm is about without having to provide all the details.
- `algorithm_service` was updated to provide new functions for CRUD operations on the algorithm database tables.
- main.py was getting large, so `routes.algorithm_router` was created to house the algorithm specific API endpoints.  Within 5 new endpoints were created for listing, retrieving, creating, updating and deleting Algorithms.
- `scripts/algorithm_config.py` was added to help CLI users easily manage the algorithms in their database.

## Checklist

## Checklist for Reviewers


src/recordlinker/linking/algorithm_service.py
```diff
@@ -1,33 +1,117 @@
...
+    :param obj: The existing algorithm object or None
+    :param commit: Commit the transaction
+    :returns: The algorithm object and a boolean indicating if it was created
     """
-    Gets a list of algorithms from the MPI database.
-    returns: list of all labels column from algorithms table
+    algo = data.model_dump()
+    passes = algo.pop("passes")
+    # use the existing Algorithm or create a new one
+    created = False if obj else True
```
cbrinson-rise8, line 77:
```suggestion
    created = not obj
```

src/recordlinker/routes/algorithm_router.py
```diff
@@ -0,0 +1,102 @@
...
+import sqlalchemy.orm as orm
+
+from recordlinker import schemas
+from recordlinker.database import get_session
+from recordlinker.linking import algorithm_service as service
+
+router = fastapi.APIRouter()
+
+
+@router.get("", status_code=fastapi.status.HTTP_200_OK)
```
cbrinson-rise8, line 21: Should we do `@router.get("/", status_code=fastapi.status.HTTP_200_OK)` just to maintain convention here?

ericbuckley: Correct me if I'm wrong, but the existing convention is no trailing slashes right?

So the current DIBBS API endpoint is, `/link-record` and not `/link-record/`.

cbrinson-rise8: True, I was basing it off `@app.get("/")` endpoint in `main.py`. But you're right, all the other endpoints don't have trailing slashes.

ericbuckley: Yeah, that's true for the health endpoint, but I think that's just a necessity because it's the root URL.

Note that I'm using a [prefix when initializing the router](https://github.com/CDCgov/RecordLinker/blob/feature/65-algorithm-config/src/recordlinker/main.py#L34).  So for this endpoint the URL will be "/algorithm".  If we add a slash here, we'll have "/algorithm/".

cbrinson-rise8: Gotcha, I overlooked that there's already `/algorithm`. Looks good!

cbrinson-rise8 (dismissed):

ericbuckley:
@cbrinson-rise8 thanks for the review.  Could you give it another look tomorrow, I had to resolve some conflicts when I rebased against main.

cbrinson-rise8 (approved):

PR #88, opened 2024-10-18
Expand Feature Comparison Support

### LINKED ISSUE #55: Expand support for feature comparisons
## Summary
NBS has requested that we add additional fields for feature comparisons.  We won't need to block on any of these features, but we do need to have the ability to compare them in the second half of the algorithm.

## Acceptance Criteria
- [x] 3 new features, that can be parsed from a FHIR payload
- [x] New tests for the PIIRecord.feature_iter method for the new features
- [x] Documentation to recap the process below

## Details / Tasks
1. Add the following values to `recordlinker.models.pii.PIIRecord`
    - SSN (should validate that its 10 digits and conform to XXX-XX-XXXX format)
    - Race (need an Enum with the [8 races identified here](https://www.hl7.org/fhir/us/core/ValueSet-omb-race-category.html))
    - Gender (need an Enum with the [5 genders identified here](https://terminology.hl7.org/5.5.0/ValueSet-gender-identity.html))
1. Add the above 3 to `recordlinker.models.pii.Feature`, and additionally add 3 more
    - Telephone
    - Suffix
    - County
1. Rename `recordlinker.models.pii.PIIRecord.field_iter` to `recordlinker.models.pii.PIIRecord.feature_iter`
1. Update `recordlinker.models.pii.PIIRecord.field_iter` to extract the 6 features above
1. Update `recordlinker.linking.fhir_record_to_pii_record` to extract the 6 features from a FHIR payload
1. Add documentation to the `docs/` directory to explain the process of adding a new Feature

## Background / Context
This is the NBS Request we received for expansion
```text
Include additional fields like race, gender, county for matching
Add the following fields as available options for blocking or matching configuration:

- Second last name
- Second middle name
- Suffix
- ID type
- ID value
- ID assigning authority
- Telephone
- Race
- County
- Gender
    - Need to determine which field in NBS to use for comparison.
    - The assumption is to use the Transgender Info field since it is a dropdown list.
    - Additional Gender field might also have gender information although it will be very difficult to use if possible since it is a text field.
```

## Dependencies
#23

cbrinson-rise8:
## Description
NBS has requested that we add additional fields for feature comparisons. We won't need to block on any of these features, but we do need to have the ability to compare them in the second half of the algorithm.

## Related Issues
closes #55 

## Additional Notes
- [x] Add the following values to `recordlinker.models.pii.PIIRecord`
    - [x] SSN (should validate that its 10 digits and conform to XXX-XX-XXXX format)
    - [x] Race (need an Enum with the [8 races identified here](https://www.hl7.org/fhir/us/core/ValueSet-omb-race-category.html))
    - [x] Gender (need an Enum with the [5 genders identified here](https://terminology.hl7.org/5.5.0/ValueSet-gender-identity.html))
- [x] Add the above 3 to `recordlinker.models.pii.Feature`, and additionally add 3 more
    - [x] Telephone
    - [x] Suffix
    - [x] County
- [x] Rename `recordlinker.models.pii.PIIRecord.field_iter` to `recordlinker.models.pii.PIIRecord.feature_iter`
- [x] Update `recordlinker.models.pii.PIIRecord.field_iter` to extract the 6 features above
- [x] Update `recordlinker.linking.fhir_record_to_pii_record` to extract the 6 features from a FHIR payload
- [x] Add documentation to the `docs/` directory to explain the process of adding a new Feature

HL7 FHIR resources
- https://build.fhir.org/patient-examples-general.json.html 
- https://hl7.org/fhir/R5/patient.html
- https://hl7.org/fhir/R5/datatypes-examples.html#HumanName
- https://www.hl7.org/fhir/us/core/StructureDefinition-us-core-race.html
- https://www.hl7.org/fhir/us/core/Patient-example.json.html

## Checklist

## Checklist for Reviewers

### DIFF
--- a/docs/process_for_adding_feature.md
+++ b/docs/process_for_adding_feature.md
@@ -1,0 +1,17 @@
+# Adding a New Feature
+
+### Add the New Feature to the `Feature` Class
+- In [src/recordlinker/schemas/pii.py](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/schemas/pii.py), add the new feature to the [Feature](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/schemas/pii.py#L12C7-L12C14) enum class.
+
+### Update the `PIIRecord` Schema
+- In the same file, modify the [PIIRecord](https://github.com/CDCgov/RecordLinker/blob/c85f555e5da91d54eb8c51e3bdf0789d1e204b2f/src/recordlinker/schemas/pii.py#L97) class to include the new feature as a field.
+- If the feature requires predefined values, create an enum to represent those values.
+
+### Modify the `PIIRecord.feature_iter` Method
+- Update the [PIIRecord.feature_iter](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/schemas/pii.py#L246) method to return the value of the new feature when it's used for comparison.
+
+### Extract the FHIR Field in `fhir_record_to_pii_record`
+- In [src/recordlinker/linking/link.py](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/linking/link.py), update the [fhir_record_to_pii_record](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/linking/link.py#L26) function to map the relevant FHIR field to the new feature in [PIIRecord](https://github.com/CDCgov/RecordLinker/blob/c85f555e5da91d54eb8c51e3bdf0789d1e204b2f/src/recordlinker/schemas/pii.py#L97).
+
+### Update the Tests
+- Add or modify unit tests to verify that the new feature is properly extracted, mapped, and compared. 
--- a/src/recordlinker/linking/link.py
+++ b/src/recordlinker/linking/link.py
@@ -33,14 +33,20 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
         "birthDate": fhir_record.get("birthDate"),
         "sex": fhir_record.get("gender"),
         "address": fhir_record.get("address", []),
-        "phone": fhir_record.get("telecom", []),
         "mrn": None,
+        "ssn": None,
+        "race": None,
+        "gender": None,
+        "telecom": fhir_record.get("telecom", []),
     }
     for identifier in fhir_record.get("identifier", []):
         for coding in identifier.get("type", {}).get("coding", []):
             if coding.get("code") == "MR":
                 val["mrn"] = identifier.get("value")
+            elif coding.get("code") == "SS":
+                val["ssn"] = identifier.get("value")
     for address in val["address"]:
+        address["county"] = address.get("district", "")
         for extension in address.get("extension", []):
             if extension.get("url") == "http://hl7.org/fhir/StructureDefinition/geolocation":
                 for coord in extension.get("extension", []):
@@ -48,6 +54,17 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
                         address["latitude"] = coord.get("valueDecimal")
                     elif coord.get("url") == "longitude":
                         address["longitude"] = coord.get("valueDecimal")
+    for extension in fhir_record.get("extension", []):
+        if extension.get("url") == "http://hl7.org/fhir/us/core/StructureDefinition/us-core-race":
+            for ext in extension.get("extension", []):
+                if ext.get("url") == "ombCategory":
+                    val["race"] = ext.get("valueCoding", {}).get("display")
+        if extension.get("url") == "http://hl7.org/fhir/StructureDefinition/individual-genderIdentity":
+            for ext in extension.get("extension", []):
+                if ext.get("url") == "value":
+                    for coding in ext.get("valueCodeableConcept", {}).get("coding", []):
+                        val["gender"] = coding.get("display")
+
     return schemas.PIIRecord(**val)
 
 
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -227,6 +227,7 @@ class PIIRecord(pydantic.BaseModel):
         """
         if value:
             val = str(value).lower().strip()
+            print(val)
             try:
                 return Race(val)
             except ValueError:
@@ -236,7 +237,7 @@ class PIIRecord(pydantic.BaseModel):
                     return Race.ASIAN
                 elif "black" in val or "african american" in val:
                     return Race.BLACK
-                elif val in "native hawaiian" in val or "pacific islander" in val:
+                elif "hawaiian" in val or "pacific islander" in val:
                     return Race.HAWAIIAN
                 return Race.OTHER
 
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -57,31 +57,31 @@ class Race(enum.Enum):
     Enum for the Race field.
     """
 
-    AMERICAN_INDIAN = "american indian or alaska native"
-    ASIAN = "asian"
-    BLACK = "black or african american"
-    HAWAIIAN = "native hawaiian or other pacific islander"
-    WHITE = "white"
-    OTHER = "other"
-    ASKED_UNKNOWN = "asked but unknown"
-    UNKNOWN = "unknown"
+    AMERICAN_INDIAN = "AMERICAN_INDIAN"
+    ASIAN = "ASIAN"
+    BLACK = "BLACK"
+    HAWAIIAN = "HAWAIIAN"
+    WHITE = "WHITE"
+    OTHER = "OTHER"
+    ASKED_UNKNOWN = "ASKED_UNKNOWN"
+    UNKNOWN = "UNKNOWN"
 
     def __str__(self):
         """
         Return the value of the enum as a string.
         """
-        return self.name
+        return self.value
     
 class Gender(enum.Enum):
     """
     Enum for the Gender field.
     """
 
-    FEMALE = "identifies as female gender (finding)"
-    MALE = "identifies as male gender (finding)"
-    NON_BINARY = "identifies as gender nonbinary"
-    ASKED_DECLINED = "asked but declined"
-    UNKNOWN = "unknown"
+    FEMALE = "FEMALE"
+    MALE = "MALE"
+    NON_BINARY = "NON_BINARY"
+    ASKED_DECLINED = "ASKED_DECLINED"
+    UNKNOWN = "UNKNOWN"
 
     def __str__(self):
         """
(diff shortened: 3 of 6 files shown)


src/recordlinker/schemas/pii.py
```diff
@@ -154,8 +200,44 @@ def parse_sex(cls, value):
...
             return Sex.UNKNOWN
+    
+    @pydantic.field_validator("ssn", mode="before")
+    def parse_ssn(cls, value):
+        """
+        Parse the ssn string 
+        """
+        if value:
+            val = str(value).strip()
+            pattern = re.compile(r"^\d{3}-\d{2}-\d{4}$")
```
ericbuckley, line 211: For increased flexibility, I think we should allow for values missing the hyphens as well.  We'd want to normalize it of course, for comparisons

docs/policies/process_for_adding_feature.md
```diff
@@ -0,0 +1,17 @@
+# Process for Adding New Feature
```
ericbuckley, line 1: Could you please move this file up one directory.

src/recordlinker/linking/link.py
```diff
@@ -33,21 +33,35 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
...
+        "ssn": None,
+        "race": None,
+        "gender": None,
+        "telecom": fhir_record.get("telecom", []),
     }
     for identifier in fhir_record.get("identifier", []):
         for coding in identifier.get("type", {}).get("coding", []):
             if coding.get("code") == "MR":
                 val["mrn"] = identifier.get("value")
+            elif coding.get("code") == "SS":
```
ericbuckley, line 46: I know this doesn't all fall on this PR, but I noticed that we don't have any unit tests for `fhir_record_to_pii_record`, would you mind adding some to the test_link.py module?

cbrinson-rise8: Yeah good callout. I thought of that too while doing this because of the code coverage percentage. I'll add some tests in

src/recordlinker/schemas/pii.py
cbrinson-rise8: Gotcha. Also do you think it makes sense to store the SSN int he `XXX-XX-XXXX` format? Otherwise we could just remove everything but the digits and store that as `XXXXXXXXX`.

ericbuckley: I had a [similar question with DOB](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/schemas/pii.py#L216), as we could optimize storage by removing the '-'.  That optimization is small though, and comes at a readability cost.  I think we should just leave them in for now.

src/recordlinker/schemas/pii.py
```diff
@@ -45,6 +52,42 @@ def __str__(self):
...
+    WHITE = "white"
+    OTHER = "other"
+    ASKED_UNKNOWN = "asked but unknown"
+    UNKNOWN = "unknown"
+
+    def __str__(self):
+        """
+        Return the value of the enum as a string.
+        """
+        return self.value
```
ericbuckley, line 73: I'm thinking maybe the keys will be a better representation of the data in the database than the values.  Its not as descriptive, but it better aligns to current values like those in the spreadsheet that Krishna sent over.  Thoughts?

cbrinson-rise8: Yeah that makes sense to me. I can do the same thing for `Gender` as well

docs/process_for_adding_feature.md
```diff
@@ -0,0 +1,17 @@
...
+- If the feature requires predefined values, create an enum to represent those values.
+
+Modify the feature_iter Method
+- Inside the PIIRecord class, update the feature_iter method to return the value of the new feature when it's used for comparison.
+
+Extract the FHIR Field in fhir_record_to_pii_record
+- Go to src.recordlinker.linking.link.py and update the fhir_record_to_pii_record function to map the relevant FHIR field to the new feature in PIIRecord.
+
+Update the Tests
+- Add or modify unit tests to verify that the new feature is properly extracted, mapped, and compared.
```
alhayward, line 17: Suggestion for readability & helpful links:

```suggestion
# Adding a New Feature

### Add the New Feature to the `Feature` Class
- In [src/recordlinker/schemas/pii.py](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/schemas/pii.py), add the new feature to the [Feature](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/schemas/pii.py#L12C7-L12C14) enum class.

### Update the `PIIRecord` Schema
- In the same file, modify the [PIIRecord](https://github.com/CDCgov/RecordLinker/blob/c85f555e5da91d54eb8c51e3bdf0789d1e204b2f/src/recordlinker/schemas/pii.py#L97) class to include the new feature as a field.
- If the feature requires predefined values, create an enum to represent those values.

### Modify the `PIIRecord.feature_iter` Method
- Update the [PIIRecord.feature_iter](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/schemas/pii.py#L246) method to return the value of the new feature when it's used for comparison.

### Extract the FHIR Field in `fhir_record_to_pii_record`
- In [src/recordlinker/linking/link.py](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/linking/link.py), update the [fhir_record_to_pii_record](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/linking/link.py#L26) function to map the relevant FHIR field to the new feature in [PIIRecord](https://github.com/CDCgov/RecordLinker/blob/c85f555e5da91d54eb8c51e3bdf0789d1e204b2f/src/recordlinker/schemas/pii.py#L97).

### Update the Tests
- Add or modify unit tests to verify that the new feature is properly extracted, mapped, and compared. 
```

src/recordlinker/schemas/pii.py
```diff
@@ -200,6 +288,28 @@ def field_iter(self, feature: Feature) -> typing.Iterator[str]:
             for name in self.name:
                 if name.family:
                     yield name.family
+        elif feature == Feature.SSN:
+            if self.ssn:
+                yield self.ssn
+        elif feature == Feature.RACE:
+            if self.race:
+                yield str(self.race)
```
alhayward, line 296: Are `race` and `gender` not already `str` values?

tests/unit/schemas/test_pii.py
```diff
@@ -88,6 +93,7 @@ def test_parse_invalid_birthdate(self):
     def test_parse_sex(self):
         record = pii.PIIRecord(sex="M")
         assert record.sex == pii.Sex.MALE
+        print(record.sex)
```
alhayward, line 96: Is this an artifact from testing?

alhayward (commented):
Nice work! Left two questions & a suggestion. I also left two clarifying [questions](https://github.com/CDCgov/RecordLinker/issues/55#issuecomment-2438628112) on the Related Issue #55.

src/recordlinker/schemas/pii.py
ericbuckley: No, they're enums

tests/unit/schemas/test_pii.py
cbrinson-rise8: yes. I'll remove that!

cbrinson-rise8:
@ericbuckley @alhayward I adjust the `race` and `gender` parsers to handle anything that doesn't map easily to just be either `unkown` or `other`. This is mainly to handle any values that aren't a part of these: https://www.hl7.org/fhir/us/core/ValueSet-omb-race-category.html 
https://terminology.hl7.org/5.5.0/ValueSet-gender-identity.html

Do y'all think this is sufficient for now? Otherwise my other thought was to just make the race field a `string` instead of an enum since there could be many different values depending on the data source.

tests/unit/linking/test_link.py
```diff
@@ -189,3 +189,109 @@ def test_enhanced_match_three(self, session, enhanced_algorithm, patients: list[
...
+    assert pii_record.address[0].city == "Boston"
+    assert pii_record.address[0].state == "Massachusetts"
+    assert pii_record.address[0].postal_code == "99999"
+    assert pii_record.address[0].county == "county"
+    assert pii_record.mrn == "1234567890"
+    assert pii_record.ssn == "111-22-3333"
+    assert pii_record.telecom[0].value == "123-456-7890"
+    assert pii_record.telecom[0].system == "phone"
+    assert str(pii_record.race) == "WHITE"
+    assert str(pii_record.gender) == "FEMALE"
```
ericbuckley, line 297: This is correct, but I'm not seeing how you mapped "Identifies as female gender (finding)" to "FEMALE".  Does the translation happen in the `fhir_record_to_pii_record` function or in the pii schema?

ericbuckley: Oh I see, it tries to parse it as FHIR data, then falls back to check for tokens in the string.  However, for if NBS sends a payload like the following to the new link PII endpoint, will it result in a value of `Race.OTHER`?

```json
{
  "race": "Black"
}
```

cbrinson-rise8: yes it would right now. I can add a couple if checks for some such as 
```
if val in ["black", "african American"]:
    return Race.BLACK
elif val in ["American indian", "alaska native"]:
    return Race.AMERICAN_INDIAN
...
...
...
return Race.OTHER
```

cbrinson-rise8: I updated the gender parser to do something similar to the above in case the gender string isn't an exact match to the enum value

ericbuckley: So a little bit of context as we move away from FHIR and open this up to supporting more formats.

FHIR is quite verbose, and while it works to model our PII structure and enums around its definition, we'll likely have to be quite a bit more flexible when parsing data from other formats (eg NBS database, eCR, elr) as they won't be as verbose in their definitions.

I think some good test cases for these parsers would be to look at the spreadsheet that Krishna sent to us and check to see if the strings are parsed accordingly.  I think the one exception would be "race=Hispanic", I believe that should just be for ethnicity.

cbrinson-rise8: Added some additional tests for the examples that Krishna sent over. And also added some if checks to `Gender` and `Race` to cover less verbose wording such as "black", "hawaiin" and other single word values.

ericbuckley: I guess I'm specifically worried about [this function](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/models/mpi.py#L100).  That is when we deserialize data from the database, are we able to reconstruct gender and race back into valid enums?  It's a little complicated so if you want to talk about this in detail later today I have time.

cbrinson-rise8: Yeah that sounds good! I'd be down to talk about it later today to make sure it would work how we want it to

cbrinson-rise8: I updated the [parse race](https://github.com/CDCgov/RecordLinker/pull/88/files#:~:text=def%20parse_race(cls%2C%20value)%3A) and parse gender functions and the corresponding enums to be non verbose and fhir based but still flexible

src/recordlinker/schemas/pii.py
```diff
@@ -154,8 +201,73 @@ def parse_sex(cls, value):
...
+            except ValueError:
+                if "female" in val:
+                    return Gender.FEMALE
+                elif "male" in val:
+                    return Gender.MALE
+                elif "nonbinary" in val:
+                    return Gender.NON_BINARY
+                elif "declined" in val or "asked" in val:
+                    return Gender.ASKED_DECLINED
+                elif "unknown" in val:
```
ericbuckley, line 266: You could probably remove this last elif, because its also the default case

tests/unit/schemas/test_pii.py
```diff
@@ -105,19 +110,113 @@ def test_parse_sex(self):
...
+        record = pii.PIIRecord(race="black or african american")
+        assert record.race == pii.Race.BLACK
+        record = pii.PIIRecord(race="native hawaiian or other pacific islander")
+        assert record.race == pii.Race.HAWAIIAN
+        record = pii.PIIRecord(race="asked but unknown")
+        assert record.race == pii.Race.ASKED_UNKNOWN
+        record = pii.PIIRecord(race="unknown")
+        assert record.race == pii.Race.UNKNOWN
+
+        # testing less verbose races
```
ericbuckley, line 136: thank you, these tests are great

ericbuckley (approved):
Thanks for all the extra tests, that really helped clear up questions I had about behavior.

ericbuckley (approved):

PR #89, opened 2024-10-18
Auto initialize algorithms in DB

### LINKED ISSUE #57: Auto initialize database with algorithms
## Summary
We don't need the complexity of alembic, or at least not yet.  What's really necessary is to build the database tables, and then load the default algorithms on startup.  This will make the process of onboarding simple in the near future.

## Acceptance Criteria
- [ ] Default algorithms should be auto created when the API server starts up

## Details / Tasks
- Remove the alembic migrations and its dependencies from pyproject.toml
- Create a new file `assets/default_algorithms.yml` and define the DIBBS Basic and DIBBS Enhanced algorithms in this file
    - Put plenty of comments in the file to describe the different elements
- In `recordlinker.database` create a new function for initializing the database, it should invoke "models.Base.metadata.create_all"
- The new function will also need to test to see if the Algorithms table has a default algorithm loaded, if so do nothing
- If there is no default algorithm, it should load the algorithms defined in `assets/default_algorithms.yml`
- If no default algorithm is defined in the yml file, it should raise an error
- It should override any existing algorithms in the database with the same name

## Background / Context
Currently, JSON is being used to store algorithms in text (or at least for our test cases).  JSON doesn't allow for comments, which limits our ability to describe the process to the future user.  We could put the documentation in another file, but typically keeping this close to the code as possible makes it more likely that the information will be used.  [YAML](https://yaml.org/) allows for comments and is generally more readable.

## Dependencies
#24

ericbuckley:
## Description
Configuring the code to optionally load in initial algorithms after the tables have been initialized.  Doing so, removes our needs now for having to use a database migration tool for the near future.

## Related Issues
closes #57 

## Additional Notes
- This is contentious, but have removed alembic.  Using alembic has required a lot of manual effort to get migrations correct, with little benefit this early in a project.  Our schema hasn't stabilized yet, and we don't have any systems running the code in production, so it's of little value.  Over time, we may need to bring this back to safely adjust production schemas, but now is not the time. 
- Created `assets/initial_algorithms.json` with the two existing DIBBs algorithms.  We can use this for our test cases, and also initializing a new database with the data, if the INITIAL_ALGORITHMS env variable is set.  This also gives customers another option for managing their algorithms, in they can write their own and set the INITIAL_ALGORITHMS env variable to load that json file.
- Capitalized `schemas.pii.Features`.  The blocking keys were capitalized, so making them both capitalized will hopefully reduce errors in writing the algorithm specs.
- Consolidated some of the assets and testing assets, as there were duplicates and there was some testing files in the top level one.

## Checklist

## Checklist for Reviewers


assets/initial_algorithms.json
```diff
@@ -0,0 +1,136 @@
+[
+    {
+        "label": "dibbs-basic",
+        "description": "The DIBBs default algorithm based on field experimentation and statistical analysis, this is a two-pass algorithm combining geographical and personal information to maximize linkage quality while minimizing false positives",
```
alhayward, line 4: I think the DIBBs Enhanced algorithm should be characterized as the default. What do you think?

From my understanding, the DIBBs Basic algorithm was created to offer users a simpler, more explainable option, but because it performs more poorly in practice because it is deterministic we opt for the DIBBs Enhanced algorithm. Probabilistic RL algorithms like the DIBBs Enhanced also seem more aligned with NBS' understanding & implementation of RL (e.g., log odds).

Also, I think the word 'deterministic' should be included in this DIBBs Basic description. As is, this description also holds for the DIBBs Enhanced algorithm, and I think the two could be differentiated more clearly. Happy to follow up with a suggestion.

assets/initial_algorithms.json
```diff
@@ -0,0 +1,136 @@
...
+                    "thresholds": {
+                        "FIRST_NAME": 0.9,
+                        "LAST_NAME": 0.9,
+                        "BIRTHDATE": 0.95,
+                        "ADDRESS": 0.9,
+                        "CITY": 0.92,
+                        "ZIP": 0.95
+                    },
+                    "true_match_threshold": 12.2,
+                    "log_odds": {
```
alhayward, line 84: How will the user update these log odds in the future? Will they need to edit this configuration file directly (or pass a new algorithm in the `INITIAL_ALGORITHMS` env var)?

src/recordlinker/linking/link.py
```diff
@@ -99,10 +99,11 @@ def compare(
     results: list[float] = []
     for field, func in funcs.items():
         # TODO: can we do this check earlier?
-        if field not in {i.value for i in schemas.Feature}:
+        feature = getattr(schemas.Feature, field, None)
+        if feature is None:
             raise ValueError(f"Invalid comparison field: {field}")
         # Evaluate the comparison function and append the result to the list
-        result: float = func(record, patient, schemas.Feature(field), **kwargs)  # type: ignore
+        result: float = func(record, patient, feature, **kwargs)  # type: ignore
```
alhayward, line 106: Nice! I think this is much more readable.

src/recordlinker/main.py
```diff
@@ -139,8 +139,9 @@ async def health_check(db_session: orm.Session = Depends(get_session)) -> Health
 
 
 # Sample requests and responses for docs
-sample_link_record_requests = utils.read_json_from_assets("sample_link_record_requests.json")
-sample_link_record_responses = utils.read_json_from_assets("sample_link_record_responses.json")
+# TODO: These assets need to be installed with the python code
```
alhayward, line 142: What does this comment mean?

tests/unit/linking/test_link.py
```diff
@@ -68,7 +63,18 @@ def test_compare_match(self):
             }
         )
 
-        algorithm_pass = models.AlgorithmPass(id=1, algorithm_id=1, blocking_keys=[1], evaluators={"first_name": "func:recordlinker.linking.matchers.feature_match_exact", "last_name": "func:recordlinker.linking.matchers.feature_match_fuzzy_string"}, rule="func:recordlinker.linking.matchers.eval_perfect_match", cluster_ratio=1.0, kwargs={})
+        algorithm_pass = models.AlgorithmPass(
+            id=1,
+            algorithm_id=1,
```
alhayward, line 68: What is the difference between the `id` and the `algorithm_id`?

alhayward (commented):
Nice work! Left a few quick questions & 1 suggestion. The justification for moving away from Alembic for now makes sense.

assets/initial_algorithms.json
ericbuckley: There are a couple of options.  1) they could use the new `/alogirhtm*` API endpoints to create and/or update Algorithms.  2) Depending on the installation, they could use the `scripts/algorithm_config.py` CLI utility to create and/or update.  3) They could create a file very similar to this one, and update the INITIAL_ALGORITHMS env variable to point to that file.  On startup, those algorithms will be loaded.

src/recordlinker/main.py
ericbuckley: Great question, so this is looking forward a bit to Dockerfile and pypi installation process.  Likely for both, but definitely for the latter, only the code in the "src/" directory is going to be installed.  So this code will fail as is because the sample files are outside of the "src/" directory.

Its not as simple as moving the `assets/` directory, because by default python is only going to install *.py files.  We'll need to update pyproject.toml to instruct the installation process to install these json files as well.

tests/unit/linking/test_link.py
ericbuckley: `id` is the primary key on the algorithm_pass table.  `algorithm_id` is the primary key on the algorithm table.

assets/initial_algorithms.json
ericbuckley: From the DIBBs teams definitions, I was leaving basic as the default because a) they said it should work for most and b) log odds needs to be trained on a STLTs data first to correctly calculate the values.

However, if you and/or the DIBBs team think we should switch the default, please feel free to make the switch.

As for the description, sure just click the "add a suggestion" button and write in a new one if you think we should change.

assets/initial_algorithms.json
```diff
@@ -0,0 +1,136 @@
...
+                        "MRN": 0.3051262572525359,
+                        "SEX": 0.7510419059643679,
+                        "STATE": 0.022376768992488694,
+                        "ZIP": 4.975031471124867
+                    }
+                }
+            }
+        ]
+    }
+]
```
cbrinson-rise8, line 136: It might be easier to reference these algorithms in tests, or other places, if we explicitly name them here.

```
basic_algorithm = { ... }
enhanced_algorithm = { ... }

algorithms = [basic_algorithm, enhanced_algorithm]
```

ericbuckley: Are you suggesting we make the top level entity in this JSON file an object rather than an array?

cbrinson-rise8: Yeah. I forgot this file is a `.json` file so ignore my suggestion above. But basically making it easier to reference these algos instead of doing 
```
 for algo in load_json_asset("initial_algorithms.json"):
        if algo["label"] == "dibbs-basic":
            return models.Algorithm.from_dict(**algo)
```

But actually you only have to load these algorithms once in `the `conf.test` file so its probably not necessary to make any change here.

ericbuckley: Yeah, I used the [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) decorator to speed it up a bit as well.  That method should only have to load the json file and parse it once, all subsequent calls to the function should just use the cached result.

assets/initial_algorithms.json
alhayward: Ah that makes sense. If it's in the DIBBs team's definitions, let's keep it that way now til further notice. I haven't discussed with them explicitly whether DIBBs Enhanced should be the default, so that may have just been my assumption.

Here's a suggestion
```suggestion
        "description": "The DIBBs Default Algorithm. Based on field experimentation and statistical analysis, this deterministic two-pass algorithm combines geographical and personal information to maximize linkage quality while minimizing false positives",
```

alhayward (approved):
LGTM!

Issue #90, opened 2024-10-18
should all Person records be used when evaluating incoming data


alhayward:

ericbuckley:
Following up on some of the discussions we had today.  As a team we decided that 
- belongingness ratio should take into account all patients within a cluster
- we should be evaluating against every patient in the cluster even if that particular patient doesn't have matching blocking criteria

For example, lets say we have 4 patients in a cluster: A, B, C and D.  If for our incoming patient Z, we block against the MRN, ZIP and SEX and only patients A, B and C match that criteria.  We still want to evaluate against all (A, B, C and D) when running the compare operations and likewise use all 4 records to calculate the belongingness ratio.

The following snippet is likely the changes needed to accomplish this
```diff
diff --git a/src/recordlinker/linking/mpi_service.py b/src/recordlinker/linking/mpi_service.py
index 8f6412b..e7b3aa0 100644
--- a/src/recordlinker/linking/mpi_service.py
+++ b/src/recordlinker/linking/mpi_service.py
@@ -22,7 +22,7 @@ def get_block_data(
     blocking keys defined in the algorithm_pass.
     """
     # Create the base query
-    base = expression.select(models.Patient.id).distinct()
+    base = expression.select(models.Patient.person_id).distinct()
 
     # Build the join criteria, we are joining the Blocking Value table
     # multiple times, once for each Blocking Key.  If a Patient record
@@ -53,7 +53,7 @@ def get_block_data(
         )
 
     # Using the subquery of unique Patient IDs, select all the Patients
-    expr = expression.select(models.Patient).where(models.Patient.id.in_(base))
+    expr = expression.select(models.Patient).where(models.Patient.person_id.in_(base))
     return session.execute(expr).scalars().all()
 ```

PR #97, opened 2024-10-25
fix(linking): fetch all Person records when calculating Belongingness…

alhayward:
… Ratio

## Description
* Fix Belongingness Ratio calculation by fetching all Patient records within the Person Clusters identified in Blocking
* Waiting to update tests until review of code changes, in case reviewers want to see the different linkage results

## Related Issues
#90 

## Checklist

## Checklist for Reviewers


src/recordlinker/linking/mpi_service.py
```diff
@@ -57,6 +57,21 @@ def get_block_data(
     return session.execute(expr).scalars().all()
 
 
+def fetch_person_records(
```
alhayward, line 60: Alternatively, we could have modified [`get_block_data()`](https://github.com/CDCgov/RecordLinker/pull/97/files#diff-ae6794ac378c7dc01b44005272a2254eabd9bca9ef6ab4169f93834c2b34ddeeL17), specifically lines [25](https://github.com/CDCgov/RecordLinker/pull/97/files#diff-ae6794ac378c7dc01b44005272a2254eabd9bca9ef6ab4169f93834c2b34ddeeL25) and [56](https://github.com/CDCgov/RecordLinker/pull/97/files#diff-ae6794ac378c7dc01b44005272a2254eabd9bca9ef6ab4169f93834c2b34ddeeL56), but I decided not to in order to preserve the definition of what "block data" is; "block data" should in fact just be those candidate Patient records identified in Blocking. However, in order to calculate Belongingness Ratio, we want to fetch all the Patient records within the Person clusters identified in Blocking. Open to feedback here, but I think modifying [`get_block_data()`](https://github.com/CDCgov/RecordLinker/pull/97/files#diff-ae6794ac378c7dc01b44005272a2254eabd9bca9ef6ab4169f93834c2b34ddeeL17) would be misleading by returning Patient records that technically were _not_ blocked on.

src/recordlinker/linking/link.py
```diff
@@ -144,7 +144,11 @@ def link_record_against_mpi(
             # block on the pii_record and the algorithm's blocking criteria, then
             # iterate over the patients, grouping them by person
             with TRACER.start_as_current_span("link.block"):
-                patients = mpi_service.get_block_data(session, record, algorithm_pass)
+                # get all candidate Patient records identified in blocking
+                candidate_patients = mpi_service.get_block_data(session, record, algorithm_pass)
+                # fetch all remaining Patient records within the Person clusters identified in
+                # blocking for Belongingess Ratio calculation
+                patients = mpi_service.fetch_person_records(session, candidate_patients)
```
ericbuckley, line 151: With this implementation, you will run an extra DB query for every blocking pass.  Have you tested the performance implications of adding these?

src/recordlinker/linking/mpi_service.py
```diff
@@ -57,6 +57,21 @@ def get_block_data(
...
+    session: orm.Session,
+    patients: typing.Sequence[models.Patient]
```
ericbuckley, line 62: You're passing in the `patients` list, but not using it in the function.  Did you mean to use on that on the base statement rather than retrieving a list of every Person in the database?

src/recordlinker/linking/mpi_service.py
cbrinson-rise8, line 60: We should write a test for this function just to make sure its working as expected

src/recordlinker/linking/link.py
alhayward: Great point, performance implications with an additional DB query was an oversight of mine.

Just pushed changes to give another pass. I looked into [Lazy Loading](https://docs.sqlalchemy.org/en/14/orm/loading_relationships.html#lazy-loading) with SQLAlchemy ORM, but wondering if returning the query expression for `candidate_patients` in `get_block_data()` and passing that to `get_person_records()` where the query is executed suffices.

src/recordlinker/linking/mpi_service.py
```diff
@@ -53,7 +53,22 @@ def get_block_data(
         )
 
     # Using the subquery of unique Patient IDs, select all the Patients
-    expr = expression.select(models.Patient).where(models.Patient.id.in_(base))
+    candidate_patients = expression.select(models.Patient).where(models.Patient.id.in_(base))
+    return candidate_patients
```
alhayward, line 57: @ericbuckley Alternatively, could we keep `get_block_data()` signatures as usual and change to the below with [`orm.lazyload()`](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#sqlalchemy.orm.lazyload)?
```suggestion
    expr = expression.select(models.Patient).where(models.Patient.id.in_(base))
    return orm.lazyload(session.execute(expr).scalars().all())
```

ericbuckley: Lazy loading isn't quite the same as lazy evaluation.  The former is about retrieving foreign key relationships, so for example if you get back a Patient object and call object.person.  The person query is lazy loaded and triggers another SQL query to retrieve the person info from the database.

Lazy evaluation is about not executing the SQL statement until you attempt to iterate over the results.  `session.execute(expr).scalars()` returns a sql result.  You can pass that around and not have the SQL statement execute, however as soon as you do something like `len([x for x in result])`, the SQL would be executed before the length was calculated.  However `session.execute(expr).scalars().all()` executes immediately, and passes a sequence object around with the results already stored in memory.  

This however doesn't work as I hoped it might in sqlalchemy.  As even if you return a SQL result, and pass that  into a new function to execute a query, its going to execute that result before executing the second query, thus resulting in two queries anyways.  Let me try to explain in a simple example

```python
result = get_persons(id=[1,2,3])
return get_patients(persons=result)
```

Take the above code, a function that returns a collection of persons that have ids 1, 2 or 3.  Then another function that returns all the patients that are within those person clusters.

Ideally the ORM would know it can create SQL for the second function like so, which would take the lazy result and create a subquery out of it, thus resulting in 1 total query
```sql
SELECT pat.*
FROM patients pat
WHERE pat.person_id IN (SELECT * FROM persons WHERE id in [1,23])
```

However, sqlalchemy isn't going to do that.  Instead of converting the result into a subquery, its going to evaluate it first, then pass it in, thus resulting in 2 executed queries.
```sql
SELECT pat.*
FROM patients pat
WHERE pat.person_id IN ([1,23])
```

The above example is contrived, and we of course don't need to call two functions if we already have a list of person ids, but hopefully it shows the differences in the SQL queries generated.

alhayward: Gotcha, I think this makes sense. In other words, the behavior we had hoped for by passing around the sql result `session.execute(expr).scalars()` without it executing is not the case with SQLAlchemy (resulting in 2 queries anyways).

In that case, does it make more sense now to take the alternative approach in just modifying the one query in `get_block_data()`?

ericbuckley (dismissed):
Looks great, thanks @alhayward

cbrinson-rise8 (dismissed):
Nice work! Looks like it just needs to have some conflicts resolved with main

ericbuckley (approved):

Issue #91, opened 2024-10-18
implement approach 1 for possible matches


alhayward:

ericbuckley:
Some ideas on what our JSON responses could look like

```json
// Possible Match
{
    "patient_reference_id": "266c775e-4b03-489b-b290-03181bfdfea2",
    "person_reference_id": null,
    "results": [
      {
        "person_reference_id": "8e5a91ce-4436-4673-9cc9-9835a59dfb92",
        "ratio": 0.85
      },
      {
        "person_reference_id": "8e5a91ce-4436-4673-9cc9-9835a59dfb93",
        "ratio": 0.75
      }
    ]
}


// Match
{
    "patient_reference_id": "266c775e-4b03-489b-b290-03181bfdfea2",
    "person_reference_id": "8e5a91ce-4436-4673-9cc9-9835a59dfb92",
    "results": [
      {
        "person_reference_id": "8e5a91ce-4436-4673-9cc9-9835a59dfb92",
        "ratio": 0.95
      }
    ]
}


// No match
{
    "patient_reference_id": "266c775e-4b03-489b-b290-03181bfdfea2",
    "person_reference_id": "8e5a91ce-4436-4673-9cc9-9835a59dfb95",
    "results": []
}
```

PR #111, opened 2024-10-30
feat(linking): implement possible match short-term

alhayward:
## Description
* Implement Possible Match using short-term approach: Apply a user-configurable range to the Belongingness Ratio threshold.
* To support Possible Match, update Link API response as follows:

**Example:** `belongingness_ratio=[0.75, 0.9]`

**No Match**
```
{
    "patient_reference_id": "5a12bf00-16f8-4c03-b980-81d3f5ff6db6",
    "person_reference_id": "b104942e-6327-4643-801a-5fb22cb621c4", # New Person ID
    "results": []
}
```

**Possible Match**
```
{
    "patient_reference_id": "c5fcfc39-a915-4b65-afc7-b681c892652e",
    "person_reference_id": None,
    "results": [
        {
            "person_reference_id": "bf27f9f2-2882-4e6f-9db9-6a916679cff7",
            "belongingness_ratio": 0.8
        }
    ]
}
```

**Match** `include_multiple_matches=false`
```
{
    "patient_reference_id": "2eb14d86-217f-45c3-a528-d038e7eadce7",
    "person_reference_id": "6abb63cb-6c31-4142-bdbf-8b2956dab1c1",
    "results": [
        {
            "person_reference_id": "6abb63cb-6c31-4142-bdbf-8b2956dab1c1",
            "belongingness_ratio": 0.98
        },
    ]
}
```

**Match** `include_multiple_matches=true`
```
{
    "patient_reference_id": "2eb14d86-217f-45c3-a528-d038e7eadce7",
    "person_reference_id": "6abb63cb-6c31-4142-bdbf-8b2956dab1c1", # Person ID with highest Belongingness Ratio
    "results": [
        {
            "person_reference_id": "6abb63cb-6c31-4142-bdbf-8b2956dab1c1",
            "belongingness_ratio": 0.98
        },
        {
            "person_reference_id": "8acf6104-d633-4516-8c7f-2c6e88a8ceae",
            "belongingness_ratio": 0.8
        },
    ]
}
```

## Related Issues
#91 

## Additional Notes
* Rename `cluster_ratio` -> `belongingness_ratio` for clarity
* Move `belongingness_ratio` definition from `AlgorithmPass` to `Algorithm` level (this was the DIBBs team's original intent)
* Add user-configurable `Algorithm.include_multiple_matches` boolean parameter to return multiple matches
    * Given `belongingness_ratio=[X, Y], if `Algorithm.include_multiple_matches=true`, return all `results` where `belongingness_ratio >= X]. Assign `person_reference_id` of Person with highest Belongingness Ratio.
* Add test coverage for
    * Possible Match result (Basic, Enhanced Algorithm)
    * New Link API response (`LinkResponse`, `LinkFhirResponse`)
    * Include multiple matches or not (`Algorithm.include_multiple_matches`)
* Add new test patient bundle assets for above coverage

## Checklist

## Checklist for Reviewers

### DIFF
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -89,7 +89,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_4"
-        echo "$RESPONSE_4" | jq -e '.prediction == match' 
+        echo "$RESPONSE_4" | jq -e '.prediction == "match"' 
         echo "$RESPONSE_4" | jq -e --arg id "$PERSON_REFERENCE_ID" '.person_reference_id == $id'     
       
         #invalid tests
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -62,7 +62,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_1"
-        echo "$RESPONSE_1" | jq -e '.is_match == false' 
+        echo "$RESPONSE_1" | jq -e '.prediction == no_match' 
 
         PERSON_REFERENCE_ID=$(echo "$RESPONSE_1" | jq -r '.person_reference_id')
 
@@ -71,7 +71,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_2"
-        echo "$RESPONSE_2" | jq -e '.is_match == true'  
+        echo "$RESPONSE_2" | jq -e '.prediction == match'  
         echo "$RESPONSE_2" | jq -e --arg id "$PERSON_REFERENCE_ID" '.person_reference_id == $id'
 
         #enhanced tests
@@ -80,7 +80,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_3"
-        echo "$RESPONSE_3" | jq -e '.is_match == false'  
+        echo "$RESPONSE_3" | jq -e '.prediction == no_match'  
 
         PERSON_REFERENCE_ID=$(echo "$RESPONSE_3" | jq -r '.person_reference_id')
 
@@ -89,7 +89,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_4"
-        echo "$RESPONSE_4" | jq -e '.is_match == true' 
+        echo "$RESPONSE_4" | jq -e '.prediction == match' 
         echo "$RESPONSE_4" | jq -e --arg id "$PERSON_REFERENCE_ID" '.person_reference_id == $id'     
       
         #invalid tests
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -62,7 +62,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_1"
-        echo "$RESPONSE_1" | jq -e '.prediction == no_match' 
+        echo "$RESPONSE_1" | jq -e '.prediction == "no_match"' 
 
         PERSON_REFERENCE_ID=$(echo "$RESPONSE_1" | jq -r '.person_reference_id')
 
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -71,7 +71,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_2"
-        echo "$RESPONSE_2" | jq -e '.prediction == match'  
+        echo "$RESPONSE_2" | jq -e '.prediction == "match"'  
         echo "$RESPONSE_2" | jq -e --arg id "$PERSON_REFERENCE_ID" '.person_reference_id == $id'
 
         #enhanced tests
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -80,7 +80,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_3"
-        echo "$RESPONSE_3" | jq -e '.prediction == no_match'  
+        echo "$RESPONSE_3" | jq -e '.prediction == "no_match"'  
 
         PERSON_REFERENCE_ID=$(echo "$RESPONSE_3" | jq -r '.person_reference_id')
 
--- a/src/recordlinker/schemas/algorithm.py
+++ b/src/recordlinker/schemas/algorithm.py
@@ -26,7 +26,6 @@ class AlgorithmPass(pydantic.BaseModel):
     blocking_keys: list[str]
     evaluators: dict[str, str]
     rule: str
-    belongingness_ratio: tuple[float, float]
     kwargs: dict[str, typing.Any] = {}
 
     @pydantic.field_validator("blocking_keys", mode="before")
@@ -67,6 +66,21 @@ class AlgorithmPass(pydantic.BaseModel):
             raise ValueError(f"Invalid matching rule: {value}")
         return value
 
+
+class Algorithm(pydantic.BaseModel):
+    """
+    The schema for an algorithm record.
+    """
+
+    model_config = pydantic.ConfigDict(from_attributes=True)
+
+    label: str = pydantic.Field(pattern=r"^[a-z0-9]+(?:-[a-z0-9]+)*$")
+    description: typing.Optional[str] = None
+    is_default: bool = False
+    include_multiple_matches: bool = True
+    belongingness_ratio: tuple[float, float]
+    passes: typing.Sequence[AlgorithmPass]
+
     @pydantic.field_validator("belongingness_ratio", mode="before")
     def validate_belongingness_ratio(cls, value):
         """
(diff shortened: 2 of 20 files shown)


src/recordlinker/schemas/algorithm.py
```diff
@@ -78,8 +77,24 @@ class Algorithm(pydantic.BaseModel):
     label: str = pydantic.Field(pattern=r"^[a-z0-9]+(?:-[a-z0-9]+)*$")
     description: typing.Optional[str] = None
     is_default: bool = False
+    include_multiple_matches: bool = True
+    belongingness_ratio: tuple[float, float]
```
ericbuckley, line 81: I recently [discovered a slightly better way to do this](https://github.com/CDCgov/RecordLinker/pull/126/files#diff-f8b75e7f2665e1eea4eee28b690b98fa9fd3b2687327a847c80df04145ecb069R40).  It won't remove the requirements for a separate validator function, as we still need to validate that the lower is less than or equal to the upper.  However, it will provide more documentation to the end user in the API on the valid range.

```python
belongingness_ratio: tuple[Annotated[float, Field(ge=0, le=1)], Annotated[float, Field(ge=0, le=1)]]
```

src/recordlinker/schemas/link.py
```diff
@@ -31,21 +31,62 @@ class LinkInput(pydantic.BaseModel):
     )
 
 
+class LinkResult(pydantic.BaseModel):
+    """
+    TODO
```
ericbuckley, line 36:
```suggestion
    Schema for linkage results to a person cluster.
```

src/recordlinker/schemas/link.py
```diff
@@ -69,23 +110,14 @@ class LinkFhirInput(pydantic.BaseModel):
...
-        description="A true value indicates that one or more existing records "
-        "matched with the provided record, and these results have been linked."
-    )
-    updated_bundle: dict = pydantic.Field(
-        description="If link_found is true, returns the FHIR bundle with updated"
-        " references to existing Personresource. If link_found is false, "
+    updated_bundle: dict | None = pydantic.Field(
+        description="If 'prediction' is 'match', returns the FHIR bundle with updated"
+        " references to existing Person resource. If 'prediction' is 'no_match', "
         "returns the FHIR bundle with a reference to a newly created "
```
ericbuckley, line 121: Can we also add that in the case of a 'possible_match', updated_bundle will be `null`.

ericbuckley: Oh never mind, I see that comment below!

src/recordlinker/schemas/link.py
```diff
@@ -69,23 +110,14 @@ class LinkFhirInput(pydantic.BaseModel):
...
-        "Person resource."
-    )
-    message: typing.Optional[str] = pydantic.Field(
-        description="An optional message in the case that the linkage endpoint did "
-        "not run successfully containing a description of the error that happened.",
-        default="",
+        "Person resource. If 'prediction' is 'possible_match', returns None."
```
ericbuckley, line 122:
```suggestion
        "Person resource. If 'prediction' is 'possible_match', returns null."
```

tests/unit/schemas/test_algorithm.py
```diff
@@ -19,15 +19,15 @@ def test_validate_blocking_keys(self):
                 blocking_keys=keys,
                 evaluators={},
                 rule="func:recordlinker.linking.matchers.eval_perfect_match",
-                cluster_ratio=0.5,
+                belongingness_ratio=[0.25, 0.5],
```
ericbuckley, line 22: Could we add a few tests to this file showing that the validators work as expected...lower has to be between 0 and 1, upper has to be between 0 and 1, lower has to be less than or equal to upper.

tests/unit/routes/test_link_router.py
```diff
@@ -421,22 +537,32 @@ def test_link_enhanced_algorithm(
...
         bundle_6["entry"] = [entry_list[5]]
         response_6 = client.post(
             "/link/fhir", json={"bundle": bundle_6, "algorithm": "dibbs-enhanced"}
         )
-        assert not response_6.json()["is_match"]
+        person_6 = response_6.json()["person_reference_id"]
+        assert response_6.json()["patient_reference_id"] and uuid.UUID(response_6.json()["patient_reference_id"])
+        assert person_6
+        assert response_6.json()["prediction"] == "no_match"
+        assert not response_6.json()["results"]
```
ericbuckley, line 565: I see lots of tests here showing "match" or "no_match", thank you for all of those!  Would it be possible here to also include a test case that highlights a "possible_match"?

.github/workflows/check_smoke_tests.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_1"
-        echo "$RESPONSE_1" | jq -e '.is_match == false' 
+        echo "$RESPONSE_1" | jq -e '.prediction == no_match'
```
ericbuckley, line 65: I think you want to compare to the string value here.
```suggestion
        echo "$RESPONSE_1" | jq -e '.prediction == "no_match"' 
```

tests/unit/routes/test_link_router.py
alhayward: I test the "possible_match" result (for [Basic](https://github.com/CDCgov/RecordLinker/pull/111/files#diff-42a8bc5fd67e453879d82298e5b7fc88e58a4ecce254e3954af3d87fc0fde518R179-R201) and [Enhanced](https://github.com/CDCgov/RecordLinker/pull/111/files#diff-42a8bc5fd67e453879d82298e5b7fc88e58a4ecce254e3954af3d87fc0fde518R230-R252)) in [test_link.py](https://github.com/CDCgov/RecordLinker/pull/111/files#diff-42a8bc5fd67e453879d82298e5b7fc88e58a4ecce254e3954af3d87fc0fde518). Is that sufficient? Or should we also include tests for this in test_link_router.py?

ericbuckley: Oh cool, thats the important one!  At this level we'd mostly be testing the if/else logic in the prediction computed field, but that function is pretty simple.  Feel free to skip if its complicated.

.github/workflows/check_smoke_tests.yml
```diff
@@ -71,7 +71,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_2"
-        echo "$RESPONSE_2" | jq -e '.is_match == true'  
+        echo "$RESPONSE_2" | jq -e '.prediction == match'
```
cbrinson-rise8, line 74:
```suggestion
        echo "$RESPONSE_2" | jq -e '.prediction == "match"'  
```

.github/workflows/check_smoke_tests.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_3"
-        echo "$RESPONSE_3" | jq -e '.is_match == false'  
+        echo "$RESPONSE_3" | jq -e '.prediction == no_match'
```
cbrinson-rise8, line 83:
```suggestion
        echo "$RESPONSE_3" | jq -e '.prediction == "no_match"'  
```

.github/workflows/check_smoke_tests.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         -H "Content-Type: application/json")
 
         echo "Response: $RESPONSE_4"
-        echo "$RESPONSE_4" | jq -e '.is_match == true' 
+        echo "$RESPONSE_4" | jq -e '.prediction == match'
```
cbrinson-rise8, line 92:
```suggestion
        echo "$RESPONSE_4" | jq -e '.prediction == "match"' 
```

.github/workflows/check_smoke_tests.yml
cbrinson-rise8: I think after changing each of these to a string the smoke tests should be good. Since `.person_reference_id` still exists on the response object that part should be the same. If any other errors come up with the smoke tests after these changes I'll take another look!

tests/unit/schemas/test_algorithm.py
```diff
@@ -71,27 +66,65 @@ def test_validate_rule(self):
...
-            cluster_ratio=0.5,
         )
+
+
+class TestAlgorithm:
+    def test_validate_belongingness_ratio(self):
+        belongingness_ratio=(0.9, 0.75)
+        with pytest.raises(pydantic.ValidationError):
+            Algorithm(
+                label="label",
```
alhayward, line 96: @ericbuckley I initially did not pass a `label` arg because we set a [default](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/schemas/algorithm.py#L78), but then I got the below error. Passing a `label` resolved it. Any ideas why the default failed?
Screenshot 2024-11-15 at 3 47 26 PM

ericbuckley: @alhayward only `description` and `is_default` have default values (the empty string and False respectively).  The `label` attribute is assigned to be a [pydantic.Field](https://docs.pydantic.dev/latest/concepts/fields/) with a regex for validation.  It only has a default value if one is explicitly set when [initializing the field](https://docs.pydantic.dev/latest/concepts/fields/#default-values).

PR #113, opened 2024-10-31
Drivers License Feature

### LINKED ISSUE #95: Feature to compare Driver's License number
## Summary
Add a new Feature to run comparisons on driver's license numbers.

## Acceptance Criteria
- [x] Add a new feature for storing and comparing `drivers_license` attribute on the PIIRecord
- [x] Add the ability to parse a driver's license number from FHIR
- [x] Test cases to show feature comparisons work

## Details / Tasks
Within a FHIR payload, the driver's license can be found in the following
```json
{
  "resourceType": "Patient",
  "id": "example",
  "identifier": [
    {
      "use": "official",
      "type": {
        "coding": [
          {
            "system": "http://terminology.hl7.org/CodeSystem/v2-0203",
            "code": "DL",
            "display": "Driver's License"
          }
        ]
      },
      "value": "D1234567",
      "assigner": {
        "display": "State DMV"
      }
    }
  ]
}
```

cbrinson-rise8:
## Description
Add a new Feature to run comparisons on driver's license numbers.

## Related Issues
closes #95 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/hl7/fhir.py
+++ b/src/recordlinker/hl7/fhir.py
@@ -27,6 +27,7 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
         "race": None,
         "gender": None,
         "telecom": fhir_record.get("telecom", []),
+        "drivers_license": None,
     }
     for identifier in fhir_record.get("identifier", []):
         for coding in identifier.get("type", {}).get("coding", []):
@@ -34,6 +35,13 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
                 val["mrn"] = identifier.get("value")
             elif coding.get("code") == "SS":
                 val["ssn"] = identifier.get("value")
+            elif coding.get("code") == "DL":
+                license_number = identifier.get("value")
+                authority = identifier.get("assigner", {}).get("identifier", {}).get("value")  # Assuming `issuer` contains authority info
+                val["drivers_license"] = {
+                    "value": license_number,
+                    "authority": authority
+                }
     for address in val["address"]:
         address["county"] = address.get("district", "")
         for extension in address.get("extension", []):
@@ -53,7 +61,7 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
                 if ext.get("url") == "value":
                     for coding in ext.get("valueCodeableConcept", {}).get("coding", []):
                         val["gender"] = coding.get("display")
-
+    
     return schemas.PIIRecord(**val)
 
 def add_person_resource(
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -29,6 +29,7 @@ class Feature(enum.Enum):
     TELEPHONE = "TELEPHONE"
     SUFFIX = "SUFFIX"
     COUNTY = "COUNTY"
+    DRIVERS_LICENSE = "DRIVERS_LICENSE"
 
     def __str__(self):
         """
@@ -133,10 +134,19 @@ class Telecom(pydantic.BaseModel):
 
     model_config = pydantic.ConfigDict(extra="allow")
 
-    value: str  # future use
+    value: str
     system: typing.Optional[str] = None
     use: typing.Optional[str] = None
 
+class DriversLicense(pydantic.BaseModel):
+    """
+    The schema for a Drivers License record
+    """
+
+    model_config = pydantic.ConfigDict(extra="allow")
+
+    value: str
+    authority: str
 
 class PIIRecord(pydantic.BaseModel):
     """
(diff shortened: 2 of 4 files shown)


tests/unit/hl7/test_fhir.py
```diff
@@ -36,6 +36,22 @@ def test_fhir_record_to_pii_record():
                         "code" : "SS"
                     }]
                 },
+            },
```
ericbuckley, line 39: I think the payload would look a bit more like the following
```
  {
    "use": "official",
    "type": {
      "coding": [
        {
          "system": "http://terminology.hl7.org/CodeSystem/v2-0203",
          "code": "DL",
          "display": "Driver's License"
        }
      ],
      "text": "Driver's License"
    },
    "system": "urn:oid:2.16.840.1.113883.3.19.3.1.8",
    "value": "D1234567",
    "assigner": {
      "display": "California Department of Motor Vehicles",
      "identifier": {
        "system": "urn:oid:2.16.840.1.113883.19.3.1.1", 
        "value": "CA"
      }
    }
  }
```

src/recordlinker/schemas/pii.py
```diff
@@ -332,6 +345,12 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
             for address in self.address:
                 if address.county:
                     yield address.county
+        elif feature == Feature.DRIVERS_LICENSE:
+            if self.drivers_license:
+                if self.drivers_license.value:
+                    yield self.drivers_license.value
+                if self.drivers_license.authority:
+                    yield self.drivers_license.authority
```
ericbuckley, line 353: Depending on the matching function that's being used, the results could be weird (particularity with the "any" matchers, as we'd just be checking to see if the ID was issued in the same state).  I wonder if it would be better to concat in this case, open to ideas though.

```python
yield f"{self.drivers_license.authority}|{self.drivers_license.value}"
```

tests/unit/hl7/test_fhir.py
ericbuckley: I think so, I don't think the display name is going to be consistent.  I think sometimes we might just see "DMV" or "Department of Motor Vehicles", in which case we won't know what state issued the license.  From the FHIR docs, display is just "Text alternative for the resource"

src/recordlinker/schemas/pii.py
cbrinson-rise8: Oh I see what you're saying. I think concatenating is a good idea. Because we essentially want to make sure the actual license value is the same and the state, not either or. So putting them together in one string makes sense to me

ericbuckley: I've been thinking about this, esp related to the triplets concept that Matthew was discussing yesterday.  In general I'm wondering if some format of `{id.type}:{id.authority}:{id.value}` will be good for making a generic id that we can compare.  Don't worry about that in this PR, but we might pivot there next week.

src/recordlinker/schemas/pii.py
```diff
@@ -133,10 +134,19 @@ class Telecom(pydantic.BaseModel):
...
 
+class DriversLicense(pydantic.BaseModel):
+    """
+    The schema for a Drivers License record
+    """
+
+    model_config = pydantic.ConfigDict(extra="allow")
+
+    value: str
+    authority: str
```
cbrinson-rise8, line 149: @ericbuckley I had the `authority` field as optional but I changed it to required assuming that the authority needs to be present for a valid comparison. Do you agree with that?

tests/unit/hl7/test_fhir.py
cbrinson-rise8: Added this payload to the test and updated the parser.

src/recordlinker/schemas/pii.py
ericbuckley: Yeah that makes sense to me

src/recordlinker/hl7/fhir.py
```diff
@@ -27,13 +27,21 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
...
     }
     for identifier in fhir_record.get("identifier", []):
         for coding in identifier.get("type", {}).get("coding", []):
             if coding.get("code") == "MR":
                 val["mrn"] = identifier.get("value")
             elif coding.get("code") == "SS":
                 val["ssn"] = identifier.get("value")
+            elif coding.get("code") == "DL":
+                license_number = identifier.get("value")
+                authority = identifier.get("assigner", {}).get("identifier", {}).get("value")  # Assuming `issuer` contains authority info
```
ericbuckley, line 40: should we add an if block here to only add a DL if both license_number and authority are non-null?

ericbuckley: oh I see, down below we only yield the value if they are defined, so I guess we don't need that if block

ericbuckley (approved):
Thanks for the quick turnaround @cbrinson-rise8 looks great.

PR #123, opened 2024-11-08
NBS Test Cases 

### LINKED ISSUE #110: implement testing script for NBS test cases
## Summary
NBS provided us a spreadsheet that defines an existing state and then a bunch of tests with expected results.  Architect a testing script of sorts, that can parse a similar spreadsheet (can just be a CSV) with seeding data and iterate through the tests with the expectations.

## Acceptance Criteria
- [x] A new script that triggers the algorithm tests
- [x] Documentation describing how to configure and run these tests

## Details / Tasks
Maybe putting this project into `tests/algorithm/` directory is a good place to isolate this testing project, but open to suggestions.
The key inputs to these tests will be the following:
- a CSV file that can be used to seed the MPI
- a CSV with test cases (basically the same format of the above), but also a column to indicate if its a match and to whom
- a JSON file with an algorithm config to use

The setup should make it easy to edit any 3 of those above files, run the test cases and examine the results.

## Background / Context
Teams thread where NBS shared the [spreadsheet](https://teams.microsoft.com/l/message/19:8d1029bc08334a5787602b4c75f1dc6e@thread.v2/1729625542966?context=%7B%22contextType%22%3A%22chat%22%7D).

cbrinson-rise8:
## Description
NBS provided us a spreadsheet that defines an existing state and then a bunch of tests with expected results. Architect a testing script of sorts, that can parse a similar spreadsheet (can just be a CSV) with seeding data and iterate through the tests with the expectations.

## Related Issues
closes #110 

## Additional Notes
For a full explanation of run instructions, directory structure, and file uses check out the [README.md](https://github.com/CDCgov/RecordLinker/pull/123/files#:~:text=%23%20Record%20Linkage%20Algorithm%20Testing)

The key inputs to these tests will be the following:

- a CSV file that can be used to seed the MPI
- a CSV with test cases (basically the same format as the above), but also a column to indicate if it is a match and to whom
- a JSON file with an algorithm config to use
The setup should make it easy to edit any 3 of those above files, run the test cases, and examine the results.

## Checklist

## Checklist for Reviewers


src/recordlinker/linking/mpi_service.py
```diff
@@ -133,7 +134,7 @@ def bulk_insert_patients(
     pat_data = [
         {
             "person_id": person.id,
-            "_data": record.to_json(prune_empty=True),
+            "_data": json.loads(record.to_json(prune_empty=True)),
```
cbrinson-rise8, line 137: @ericbuckley small bug I found in the `bulk_insert_patients` function. `_data` was being stored as a string instead of a `dict`. This caused a type error to be thrown whenever an incoming record was being compared to a patient since this line `self._record = pii.PIIRecord.model_construct(**(self._data or {}))` requires `self._data` to be a dict.

cbrinson-rise8: fixed in a different PR and rebased

tests/algorithm/scripts/send_test_records.py
```diff
@@ -0,0 +1,59 @@
+import pandas as pd
+import requests
+from helpers import dict_to_pii
+
+
+def send_test_records(test_csv, algorithm_name, api_url):
+    df = pd.read_csv(test_csv)
```
ericbuckley, line 7: Is pandas only necessary to read and write csv files?  If so, there is a standard lib for that https://docs.python.org/3/library/csv.html

tests/algorithm/scripts/run_test.py
```diff
@@ -0,0 +1,34 @@
+import os
```
ericbuckley, line 1:
```suggestion
#!/usr/bin/env python3

import os
```

tests/algorithm/README.md
```diff
@@ -0,0 +1,93 @@
...
+    `/configurations/algorithm_configuration.json`
+
+    Edit the configuration file to tune the algorithm parameters
+
+## Running Algorithm Tests
+
+1. Run the tests
+
+    ```bash
+    docker compose --profile algo-test run --rm algo-test-runner python scripts/run_test.py
```
ericbuckley, line 45:
```suggestion
    docker compose run --rm algo-test-runner scripts/run_test.py
```

tests/algorithm/scripts/run_test.py
ericbuckley: Will also need to run `chmod +x` on this file

tests/algorithm/results/output.csv
```diff
@@ -0,0 +1,85 @@
+Test Case #,Expected Result,Match Result,Error
```
ericbuckley, line 1: Did you mean to commit this results file?

ericbuckley:
@cbrinson-rise8 I was going through the README and ran the default tests.  I noticed the all of the expected results in the output were either a failure or a match, which is different from the results.csv file you committed.  Any ideas why my results would have been different from the ones in the PR?

tests/algorithm/scripts/run_test.py
cbrinson-rise8: Added and also updated the readme to call out the `chmod +x` command

tests/algorithm/results/output.csv
cbrinson-rise8: no I didn't. I'll remove it from the pr

tests/algorithm/scripts/send_test_records.py
cbrinson-rise8: Yeah. I will go ahead and switch it to that default reader. Removing it should make the build faster too.

tests/algorithm/README.md
```diff
@@ -0,0 +1,99 @@
+# Record Linkage Algorithm Testing
+
+This repository contains a project to test the effectiveness of the RecordLinker algorithm.
```
alhayward, line 3:
```suggestion
This repository contains a project to evaluate the match accuracy performance of the RecordLinker algorithm.
```

tests/algorithm/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Directory Structure
+
+- `/`: Contains the `.env` file and `Dockerfile` to build
+- `configurations/`: Contains the configuration file for the algorithm tests
+- `data/`: Contains the data `.csv` files used for the algorithm tests (seed file and test file)
+- `results/`: Contains the results of the algorithm tests
+- `scripts/`: Contains the scripts to run the algorithm tests
+
+## Steup
```
alhayward, line 20:
```suggestion
## Setup
```

tests/algorithm/README.md
```diff
@@ -0,0 +1,99 @@
...
+    Run the test
+    ```bash
+    docker compose run --rm algo-test-runner scripts/run_test.py
+    ```
+
+2. Analyze the results
+
+    The results of the algorithm tests will be available in the `results/output.csv` file.
+
+    The results will be in a csv formatted file withthe following columns: `Test Case #`, `Expected Result`, `Match Results`, and `Error`.
```
alhayward, line 58:
```suggestion
    The results will be in a CSV formatted file with the following columns: `Test Case #`, `Expected Result`, `Match Results`, and `Error`.
```

alhayward: We may want to edit these column names over time to align with naming conventions of performance metrics used for evaluating classification models, like the Record Linkage algorithm (3 classes: `match`, `no_match`, `possible_match`). E.g. `True Label` instead of `Expected Result` and `Prediction` instead of `Match Results`.

tests/algorithm/configurations/algorithm_configuration.json
```diff
@@ -0,0 +1,66 @@
...
+    "belongingness_ratio": [0.75, 0.9],
+    "passes": [
+        {
+            "blocking_keys": [
+                "BIRTHDATE"
+            ],
+            "evaluators": [
+                    {
+                        "feature": "FIRST_NAME",
+                        "func": "func:recordlinker.linking.matchers.feature_match_fuzzy_string"
```
alhayward, line 15: Depending on when this gets merged, you may need to pull the latest changes from main with the new names of the feature matching/evaluator funcs, if it's after #137 is merged.

tests/algorithm/data/seed_data.csv
```diff
@@ -0,0 +1,1001 @@
+Match Id,ID,BIRTHDATE,FIRST,LAST,SUFFIX,MAIDEN,RACE,ETHNICITY,GENDER,ADDRESS,CITY,STATE,COUNTY,ZIP,SSN,
```
alhayward, line 1: Prob fine since this is synthetic data, but do we need approval from NBS to store this on our open-source GitHub repo? CC: @ericbuckley

tests/algorithm/scripts/seed_db.py
```diff
@@ -0,0 +1,43 @@
+import csv
+
+import requests
+from helpers import dict_to_pii
+
+
+def seed_database(csv_file, api_url):
+    MAX_CLUSTERS = 100
```
alhayward, line 8: What is the purpose of constraining the number of clusters to 100?

tests/algorithm/scripts/send_test_records.py
```diff
@@ -0,0 +1,66 @@
+import csv
+
+import requests
+from helpers import dict_to_pii
+
+
+def send_test_records(test_csv, algorithm_name, api_url):
+    output_data = []
+
+    print("Sending test records to the API...")
```
alhayward, line 10: I assume these messages get printed to the CLI? What's the difference between printing vs. logging in this case?

tests/algorithm/scripts/send_test_records.py
```diff
@@ -0,0 +1,66 @@
...
+
+            pii_record = dict_to_pii(record_data)
+            
+            response = send_record(pii_record, algorithm_name, api_url)
+
+            if response:
+                output_row = {
+                    "Test Case #": match_info['test_case_number'],
+                    "Expected Result": match_info['should_match'],
+                    "Match Result": response['prediction'],
```
alhayward, line 35: Our valid `prediction` values of `"match"`, `"no_match"`, and `"possible_match"` don't exactly align with those in the `Expected Result` column of the NBS test cases. Is there transformation somewhere, or are we expecting our valid values in this column? If the latter, is validation somewhere needed?

For context, having the same set of values in the `Expected Result` and `Match Result` columns of the `output.csv` is very helpful for calculating match accuracy performance metrics down the line (e.g., False Positives, False Negatives, Sensitivity, Specificity, etc.). If there are transformations to results needed, I think that should live somewhere in this testing framework and outside of the calculations for these metrics.

E.g.,`"Should Be a Match"` in NBS test cases vs. `"match"` in Record Linkage algorithm prediction, `"Should Be a Match?"` vs. `"possible_match"`, etc.

tests/algorithm/scripts/send_test_records.py
```diff
@@ -0,0 +1,66 @@
...
+                    "Error": ""
+                }
+            else:
+                output_row = {
+                    "Test Case #": match_info['test_case_number'],
+                    "Expected Result": match_info['should_match'],
+                    "Match Result": "Failed",
```
alhayward, line 42: What would cause this `"Failed"` result?

Would we want to error out in the testing framework somewhere earlier on? Not sure how long this kind of testing would take to run, but I'm imaging the case when it takes a while only to return many `"Failed to link record"` results due to a formatting error or something easily fixed; the user would only become aware of that after the testing finished running.

alhayward (commented):
Thanks for this work! Excited that we'll have the functionality to test results, which paves the way for calculating match accuracy performance metrics!

Left several questions and a few suggestions.

tests/algorithm/scripts/seed_db.py
cbrinson-rise8: The `/seed` api endpoint has a hardcoded limit of 100 clusters to help users to keep requests a reasonable size (around 1 MB). `MAX_CLUSTERS` here just enforces that 100 cluster per batch on the sending side.

tests/algorithm/scripts/send_test_records.py
cbrinson-rise8: The print statements here are meant to give the user an idea of what is happening while they are waiting for the tests to run. It also will print out any error points along the way. Such as a failed batch seeding or records that have errors being linked. All just informational for the user during the testing process. Logging would ideally be more structured and provide more context with each message.

tests/algorithm/scripts/send_test_records.py
cbrinson-rise8: The failed result happens when a record is sent to be linked via the `/link` endpoint but returns some sort of error. Could be a couple of causes but the most likely cause is just bad data. Invalid formatting or too many characters are common ones. I could print more information related to the error in the output file or in the terminal. My only thought with erroring out earlier is that the test may almost be complete and then suddenly get stopped. My thought is that if the user sees and error in the terminal then they can stop the test if they want but otherwise the program fails that test and moves on.

alhayward: I think printing more information related to the error would be helpful, but I think this is a good discussion that brings up the question of: is the purpose of this testing framework to test the functionality of Record Linker (e.g., do not pass invalid patient records), or the match accuracy of the algorithm it uses?

I believe the latter, although allowing `"Failed to link record"` results as described above would be testing more of the functionality of Record Linker, I think. I think the goal of this kind of match accuracy testing is to, once the tests are complete, have an idea about how accurate the current algorithm is (not whether there were data formatting issues). Calculating match accuracy metrics like Sensitivity, Specificity, etc. is out of scope for this PR (I believe I'll be working on that as a fast-follow afterwards), but doing so will require "clean" prediction results (`"match"`, `"no_match"`, `"possible_match"`). It will be difficult to interpret and calculate these metrics when we have results like `"Failed to link result"`. For example, would we just exclude those results in the calculations? If so, we may report flawed accuracy metrics (what if the patient records with `"Failed to link result"` would have been accurately linked, but the user just had a data formatting issue?).

Interested in what you both think regarding if we should standardize the`"Expected Result"` and `"Match Result"` values here. IMO, it would be better to standardize these values in the testing framework rather than when calculating Sensitivity, Specificity, etc. CC: @ericbuckley

cbrinson-rise8: Yeah, I agree that the goal is not to fix errors in the user's data but rather to analyze the results of the algorithm configuration on the user's data with their test cases. I'm thinking the simple "Failed to link record" or something like that would be all we need right now since we aren't calculating any metrics yet as you said. And I know NBS even said they had some test cases with intentionally bad data that are expected to fail. 

Right now the result file shows "failed to link record" as well as a short error description. Do we think that's too much or too little info for the users at this point? Also if either of y'all have word smithing on that then feel free to let me know. this was just the first thing I thought of.

@alhayward @ericbuckley

ericbuckley: @alhayward, Cason and I spoke about this after standup today and decided that we should do the following.

a) Distinguish between server and user errors, the latter can be returned to the user in the match results, but the former should cause the program to halt immediately (since its not expected).
b) User errors ([a 422](https://cdcgov.github.io/RecordLinker/api-docs.html#tag/link/operation/link_piirecord_link_post)), will show the problem in the detail attribute of the payload, we can extract this information and put it into the Error column for the user to see.

ericbuckley:
@cbrinson-rise8 I think pretty much all of my previous questions have been resolved.  I did push a small change to [update some language in the README and the sample tests cases](https://github.com/CDCgov/RecordLinker/pull/123/commits/79556bb997a53d8400e5d222a89ac933067ceed0), please take a look to make sure you agree with those.

My only remaining thought is I think we should remove tests/algorithm/results/output.csv.  That's a generated file and not source, so I think it is best to remove that.  Additionally, we may want to add a rule to the .gitignore file, so future devs don't accidentally commit that file.  Thoughts?

ericbuckley:
@cbrinson-rise8 FYI, I also rebased against main and adjusted the algorithm_configuration.json accordingly. https://github.com/CDCgov/RecordLinker/pull/123/commits/8e1575d536926dc1c77dd15bb6760697df7986bc

ericbuckley (approved):

PR #124, opened 2024-11-08
seed API

### LINKED ISSUE #122: /seed API endpoint
## Summary
Create a new API endpoint, `/seed`, that given a batch of patient person records will insert records into the mpi tables and return the newly formed reference_ids.

## Acceptance Criteria
- [ ] New schemas to support the input/output for seeding
- [ ] The new API endpoint
- [ ] Integration tests for the new API endpoint

## Details / Tasks
Rather than having a flat list of patient data, we probably want to group records into sets to indicate that all of the incoming records in a set are in the same person cluster.
Should we have additional API endpoints for more adjustments and/or deleting data?

ericbuckley:
## Description
Adding a new API endpoint, `/seed`, to control initializing the MPI tables with data.  `POST /seed` is used to insert person/patient clusters into the MPI and `DELETE /seed` is used to reset the MPI.

## Related Issues
closes #122

## Additional Notes
- Updating the unit test CI job to combine all the coverage reports from the matrix of tests before uploading to codecov.  This changeset adds some features not compatible with MySQL, thus some test cases are skipped depending on the database dialect.
- adding `mpi_service.bulk_insert_patients`, essentially the same functionality of insert_patient, but is faster when attempting to insert more than 1 patient.
- [MySQL doesn't support all the bulk insert features](https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#getting-new-objects-with-returning), so it will be slower than the other dialects when using /seed

## Checklist

## Checklist for Reviewers


src/recordlinker/schemas/seed.py
```diff
@@ -0,0 +1,41 @@
...
+
+import pydantic
+
+from recordlinker.schemas.mpi import PatientRef
+from recordlinker.schemas.pii import PIIRecord
+
+CLUSTER_MAX_SIZE = 100
+# NOTE: the number of records in each cluster can vary, but assuming we have
+# less than ~25 records per person, we should be able to fit 100 persons in a
+# 1MB request.  This is a reasonable limit for a single request.
```
cbrinson-rise8, line 11: Should there also be a `RECORDS_MAX_SIZE` to enforce the 25 record limit per cluster assumption?

src/recordlinker/routes/seed_router.py
```diff
@@ -0,0 +1,92 @@
...
+from recordlinker.database import get_session
+from recordlinker.linking import mpi_service as service
+
+router = fastapi.APIRouter()
+
+
+@router.post("", summary="Batch seed records", status_code=fastapi.status.HTTP_201_CREATED)
+def batch(
+    data: schemas.ClusterGroup, session: orm.Session = fastapi.Depends(get_session)
+) -> schemas.PersonGroup:
```
cbrinson-rise8, line 25: Thikning of using this endpoint to seed the database with a list of PIIRecords. Would I need to structure each `PIIRecord` in its own cluster when sending the request?

src/recordlinker/schemas/seed.py
ericbuckley: This is a tricky one.

The 1MB limit isn't a hard rule. Many web servers can handle, or be configured to handle, payloads much larger than 1MB.  However, most are configured with some limit. I used nginx as a baseline here, since it's a very popular web server to put in front of application servers, and its [default value is 1MB](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size).

This API endpoint will be able to process more than that.  I'm sure it could probably handle a 5MB payload with no problem (assuming there is enough available RAM to use).  However, if the user were to send a really large payload here, and the web server fronting this application had a smaller limit, the error would happen before RL was even called.  As nginx or Apache would raise an error when it receives a body that is too large, thus this check wouldn't even happen.

More than anything, this kind of acts as documentation / guidance, on how the user should consider using this endpoint.  The array length limits really won't enforce anything on size.  It's conceivable, albeit unlikely, for just a single PII document to pass the 1MB threshold, say if they had 10000 name values or 10000 addresses or something absurd like that.  Nonsense, of course, but possible.

It might make more sense for us to just put the threshold on body size, but my worry is that would be harder for the user to control when sending requests to this API.  It just much easier for them to use some sort of page size limit.

cbrinson-rise8: Yeah you're right. Either way its not actually enforcing a size (like if the records are huge or there are a ton of them etc). I guess it makes sense to not add a ~25 patient limit to a cluster. I think the cluster limit you already have is a good idea if the goal is mainly guidance. I'm fine keeping it like this.

cbrinson-rise8: We don't really have a body size limit on any other endpoint either so I'm good with how it is

src/recordlinker/routes/seed_router.py
cbrinson-rise8: Yeah that will be easy enough to wrap the patients with a cluster and then a cluster group with all of them. Thanks!

cbrinson-rise8 (approved):

Issue #109, opened 2024-10-30
Adjust matching function names to better represent functionality


ericbuckley:
## Summary
Rename all of our functions in `recordlinker.linking.matchers` to better align with their functionality.

## Acceptance Criteria
- [x] Renaming the functions as directed in the comments of this issue
- [x] Updates to the intial_algorithms.json to match the new names
- [x] Updates in the test cases to account for name changes
- [x] Update documentation to any references to these in docs

## Details
Existing Eval Names:
- eval_perfect_match
- eval_log_odds_cutoff

Existing Matching Names:
- feature_match_any
- feature_match_exact (maybe "feature_match_all"?)
- feature_match_fuzzy_string (maybe "feature_match_fuzzy_any"?)
- feature_match_log_odds_fuzzy_compare

ericbuckley:
@alhayward do you have any name recommendations for the above function?

PR #126, opened 2024-11-14
Add additional type checks when editing an Algorithm

### LINKED ISSUE #118: use additional schemas for POST /algorithm
## Summary
Update the algorithm schemas to use pydantic schemas for blocking keys, features and evaluator functions so the users of the API docs know exactly what values are acceptable.

## Acceptance Criteria
- [ ] The redoc pages should list all the available blocking keys, features and functions on the create algorithm endpoint

ericbuckley:
## Description
Updating the way evaluators are stored on passes, so the keys are always the same.  Then adding more enums to the `AlgorithmPass` schema so we can correctly validate functions, blocking keys, features, cluster ratio and kwargs.

## Related Issues
closes #118

## Additional Notes
The code that validates rule and feature functions are of the right type was removed and replaced with something simpler.  Now all rule and feature function options are [defined as enum strings](https://github.com/CDCgov/RecordLinker/pull/126/files#diff-d3b43dc0de75a2ed734b64327582b83ee18b097049068b88384bc9fc8b424ba3R23-R52), so input is easily validated by checking to see if it matches those strings.  To verify that the strings themselves, written in the enums are valid, [unit tests were added](https://github.com/CDCgov/RecordLinker/pull/126/files#diff-4a54e1d072ebf16134d0bd906e7b527dfb0d6f06559693cc6ed99fb6c5cd7e7fR18-R43) (it will only be validated during builds, never at runtime).

Updated the site docs with some missing values.

## Checklist

## Checklist for Reviewers


cbrinson-rise8 (approved):

src/recordlinker/linking/matchers.py
```diff
@@ -18,10 +19,54 @@
...
-# A Callable type for comparing a feature on two records
-FEATURE_COMPARE_FUNC = typing.Callable[[PIIRecord, Patient, Feature, typing.Any], float]
-# A Callable type for evaluating whether a set of feature comparisons constitutes a match
-MATCH_RULE_FUNC = typing.Callable[[list[float], typing.Any], bool]
+
+class RuleFunc(enum.Enum):
+    """
+    Enum for the different types of match rules that can be used for patient
+    matching. This is the universe of all possible match rules that a user can
+    choose from when configuring their algorithm.  When data is loaded into the
```
alhayward, line 27: I like that we constrain the match rules that the user can select based on those in the default Basic and Enhanced algos (i.e., that showed highest match accuracy based on performance testing and lit review)

src/recordlinker/linking/matchers.py
```diff
@@ -18,10 +19,54 @@
...
+    AlgorithmPass schema. This is the universe of all possible keyword arguments
+    that a user can choose from when configuring their algorithm.  When data is
+    loaded into the MPI, all possible AvailableKwargs will be created for the
+    defined keyword arguments. However, only a subset will be used in matching,
+    based on the configuration of the algorithm.
+    """
+
+    SIMILARITY_MEASURE = "similarity_measure"
+    THRESHOLD = "threshold"
+    THRESHOLDS = "thresholds"
```
alhayward, line 67: What is the difference between these kwargs? If they are thresholds for different parameters, maybe we should distinguish them by name here. I can also include this in the renaming ticket #109

src/recordlinker/models/algorithm.py
```diff
@@ -16,6 +17,16 @@
 LOGGER = logging.getLogger(__name__)
 
 
+@dataclasses.dataclass
+class BoundEvaluator:
+    """
+    The schema for a bound evaluator record.
```
alhayward, line 23: What is meant by "bound" evaluator? Does this refer to a constrained set of evaluator functions as above, or to "bound" as in str -> Python func?

alhayward (approved):
LGTM, left 2 questions

src/recordlinker/linking/matchers.py
ericbuckley: This is a [holdover from the original implementation](https://github.com/CDCgov/phdi/blob/main/containers/record-linkage/app/linkage/link.py#L1021-L1025).  From what I can tell `threshold` points to a single float value that is to be used for all features, `thresholds` points to a mapping of features-> floats and used for feature specific thresholds.

In general, I'd like to rethink everything about kwargs as I think it will be difficult for users to remember when they have to specify them, and what values they need as they use different functions.  There is also the chance that some users might want different values in a pass depending on which evaluator step they are in.

src/recordlinker/models/algorithm.py
ericbuckley: It just means bound to a Callable.  An `Evaluator` has two attributes, a feature (string) and a func (string).  A `BoundEvaluator` has the same two attributes, but the func is a Callable and not a string.

PR #134, opened 2024-11-15
fix misleading matcher names

### LINKED ISSUE #128: rename TELEPHONE feature to TELECOM
## Summary
The telecom input for both `/link` and `/link/fhir` can include[ more than a phone number](https://www.hl7.org/fhir/valueset-contact-point-system.html).  Update the name of the feature to `TELECOM` so users are aware this can match on more than just a phone number.

### LINKED ISSUE #129: Feature and Blocking Key for GIVEN_NAME
## Summary
Distinctions should be made between FIRST_NAME and GIVEN_NAME, as the latter includes middle.  The Blocking Key and Feature matcher for FIRST_NAME currently has the desired behavior of GIVEN_NAME.  We should correct this behavior and add a new value for blocking and matching on GIVEN_NAME.

## Acceptance Criteria
- [ ] Update the FIRST_NAME blocking and matching to only work on the **first** element of a given name
- [ ] Create a GIVEN_NAME blocking key that creates values for all given name parts
- [ ] Create a GIVEN_NAME feature matcher that creates values for all given name parts
- [ ] Create tests to show the differences for both
- [ ] Update the reference.md site docs to explain the differences between the two

ericbuckley:
## Description
Renaming `Feature.TELEPHONE` to `Feature.TELECOM`, as that more accurately describes the matching behavior.  Adding` Feature.GIVEN_NAME` and updating `Feature.FIRST_NAME` to only use the first element of a given name.

## Related Issues
closes #128
closes #129 

## Additional Notes
- TELEPHONE was misleading because the telecom field could collect emails, fax numbers and other contact points, that would be included in the matches.  Renaming this to TELECOM better explains to the user what we're matching on.
- FIRST_NAME was misleading, as it was including the middle names in its matches as well.

## Checklist

## Checklist for Reviewers

### DIFF
--- a/docs/site/reference.md
+++ b/docs/site/reference.md
@@ -67,9 +67,9 @@ linkage evaluation phase. The following features are supported:
 
 :   The patient's county.
 
-`TELEPHONE`
+`TELECOM`
 
-:   The patient's telephone number.
+:   The patient's phone, email, fax, or other contact information.
 
 
 ### Blocking Key Types
--- a/docs/site/reference.md
+++ b/docs/site/reference.md
@@ -35,9 +35,13 @@ linkage evaluation phase. The following features are supported:
 
 :   The patient's race in the format of "AMERICAN_INDIAN", "ASIAN", "BLACK", "HAWAIIAN", "WHITE", "OTHER", "ASKED_UNKNOWN" or "UNKNOWN".
 
-`FIRST_NAME`
+`GIVEN_NAME`
 
 :   The patient's given name, this includes first and middle names.
+
+`FIRST_NAME`
+
+:   The patient's first name.
 
 `LAST_NAME`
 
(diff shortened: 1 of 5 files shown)


src/recordlinker/schemas/pii.py
```diff
@@ -323,11 +324,17 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
...
-        elif feature == Feature.FIRST_NAME:
+        elif feature == Feature.GIVEN_NAME:
             for name in self.name:
                 for given in name.given:
                     if given:
                         yield given
+        elif feature == Feature.FIRST_NAME:
+            for name in self.name:
+                # We only want the first given name for comparison
+                for given in name.given[0:1]:
```
cbrinson-rise8, line 335: Are we safe to assume that the first name of a patient will always be the first index of the `name.given` array? I guess it doesn't matter because users could choose to check the full `name.given` array by using the `GIVEN_NAME` feature if they really wanted to.

ericbuckley: I think that is a fine assumption, as long the collection remains a list as those or orderable (opposed to something like a set, which is not).  Are you thinking that this could be an issue in the FHIR payloads we parse?

docs/site/reference.md
```diff
@@ -67,9 +71,9 @@ linkage evaluation phase. The following features are supported:
 
 :   The patient's county.
 
-`TELEPHONE`
+`TELECOM`
 
-:   The patient's telephone number.
+:   The patient's phone, email, fax, or other contact information.
```
alhayward, line 76: Awesome that we can support email as a matching feature now!

alhayward (approved):
LGMT!

docs/site/reference.md
ericbuckley: I have a feeling NBS will want separate features for matching on phone vs email, but I'd like to address that in the future.

src/recordlinker/schemas/pii.py
cbrinson-rise8: Okay cool, the fact that the list is orderable makes sense to me as I was thinking which spot in the `given` array would the most recent first name be.  And no, I think it should be fine with FHIR since the given is always a list

PR #135, opened 2024-11-19
add logging to linkage process and consolidate FHIR link routes

### LINKED ISSUE #130: consolidate /link/fhir and /link/dibbs
## Summary
`/link/fhir` & `/link/dibb` do the same thing, but return a different response.  As we move towards a new response structure in #91, there is no longer a need to differ between the response structures.  These endpoints should be consolidated into one, `/link/fhir`

## Acceptance Criteria
- [ ] Add the `updated_bundle` attribute to the response of `/link/fhir`
- [ ] Update unit, performance and smoke tests to use the `/link/fhir` in place of `/link/dibbs`
- [ ] Delete the `/link/dibbs` endpoint
- [ ] Update API docs to indicate that the `/link/fhir` API can be used within the DIBBs pipeline

## Dependencies
#91

### LINKED ISSUE #131: logging for linking process
## Summary
Add additional logging statements to the `recordlinker.linking.link` module to capture key data points of the process.  Ideally we should capture enough information to understand why a match, no match or possible match happened.

## Acceptance Criteria
- [ ] Add new logging info/debug calls to link.py

## Details / Tasks
- At the pass level
  - Capture information on the blocking call
  - Capture information on the calculated belongingness ratio for each cluster evaluated
  - Possibly, capture information on the comparison results (might be too noisy)?
- Capture information on the linkage results

## Background / Context
Logging context will be important here.  The `extra` kwarg to the logging methods can be used to pass a dict of information about the context of the event.  Using this will make it much easier to analyze the records when using a log analysis tool like Splunk.

## Dependencies
#91

ericbuckley:
## Description
Add logging to link.py to capture details of linkage process.

## Related Issues
closes #131 
closes #130 

## Additional Notes
- refactored link_router to calculate the "prediction" result in link.py, as logging that value will be valuable.
- cleanup of try/except blocks in link_router to only return 422 on missing algorithm or bad FHIR bundle.
- consolidated /link/fhir and /link/dibbs into one endpoint
- updating splunk to send channel header

## Checklist

## Checklist for Reviewers


src/recordlinker/linking/link.py
```diff
@@ -113,36 +114,58 @@ def link_record_against_mpi(
                                 matched_count += 1
                     # calculate the match ratio for this person cluster
                     belongingness_ratio = matched_count / len(patients)
+                    LOGGER.info(
+                        "cluster belongingness",
+                        extra={
+                            "ratio": belongingness_ratio,
```
alhayward, line 120:
```suggestion
                            "belongingness_ratio": belongingness_ratio,
```

src/recordlinker/linking/link.py
```diff
@@ -113,36 +114,58 @@ def link_record_against_mpi(
...
+                            "person.reference_id": person.reference_id,
+                            "matched": matched_count,
+                            "total": len(patients),
+                            "algorithm.ratio_lower": belongingness_ratio_lower_bound,
```
alhayward, line 124:
```suggestion
                            "algorithm.belongingness_ratio_lower": belongingness_ratio_lower_bound,
```

src/recordlinker/linking/link.py
```diff
@@ -113,36 +114,58 @@ def link_record_against_mpi(
...
+                            "algorithm.ratio_upper": belongingness_ratio_upper_bound,
```
alhayward, line 125:
```suggestion
                            "algorithm.belongingness_ratio_upper": belongingness_ratio_upper_bound,
```

src/recordlinker/linking/link.py
```diff
@@ -113,36 +114,58 @@ def link_record_against_mpi(
...
         )
 
+    LOGGER.info(
+        "link results",
+        extra={
+            "person.reference_id": matched_person and matched_person.reference_id,
+            "patient.reference_id": patient.reference_id,
+            "result.prediction": prediction,
+            "result.count": len(results),
+            "result.scored_count": len(scores),
```
alhayward, line 166: What is the purpose of logging both `len(results)` and `len(scores)`?

src/recordlinker/linking/link.py
```diff
@@ -113,36 +114,58 @@ def link_record_against_mpi(
                                 matched_count += 1
```
alhayward, line 114: Do we plan on logging any of the feature comparison results (e.g., similarity score, log odds score, log odds ratio, fuzzy threshold), as indicated in #131?

I think it would be helpful for explainability purposes (why was this predicted a `match` or not?). I found myself often logging these results manually while working on past tickets to better understand why a Patient record was predicted a `match` or not.

alhayward (commented):
I think logging will help observability and explainability! Left a few questions/suggestions

src/recordlinker/linking/link.py
ericbuckley: `len(results)` is the number of results we return to the user, so its showing us how often we have more than 1 record in the upper threshold. Although as I write this, thats only true if the `include_multiple_matches == true` , so maybe I need to revise this to work in all cases.  My thought is that if we often get results where this number is above 1, clustering is suboptimal. 

`len(scores)` is telling us how many results are above the lower threshold.  I was thinking this would somehow convey to us how well the blocking queries are being set up (or are we too limiting, or are we including too many records), but again there may be a mistake here in that we aren't including results that are below the threshold.  So its not exact, but maybe a proxy for how well the blocking queries are performing?

Thanks for questioning this, we can probably to better with these metrics.

src/recordlinker/linking/link.py
ericbuckley: Let me add something now just so we have it.  We can run some tests, and if its still too much noise we can downgrade it to a debug statements so its not always present.

ericbuckley: I added some logging to the compare function for this

https://github.com/CDCgov/RecordLinker/pull/135/files#diff-7bc8ec967134462397574bcf3aa1c2be6e37f22c69ce36cbb36fbcbce8b1136cR64

src/recordlinker/linking/link.py
alhayward: >My thought is that if we often get results where this number is above 1, clustering is suboptimal.

I think this depends on what the upper threshold is (in addition to whether `include_multiple_matches=true`, as you said). I'm not sure if this is a valid interpretation in all cases. For example: 

1. If `belongingness_ratio=(0.9, 0.92)` and `include_multiple_matches=true`, then that might be a fair assumption (interpretation of `len(results)` > 1: multiple matches were found where >= 90% Patient records match the incoming Patient record - should these two clusters be the same Person? Can point to poor-quality clusters)
2. If `belongingness_ratio=(0.9, 0.92)` and `include_multiple_matches=false`, `len(results)` will always be 1, which is not meaningful information.
3. If `belongingness_ratio=(0.75, 0.9)` and `include_multiple_matches=true`, then I'm not sure this necessarily points to poor clusters. This would just mean that all results where >= 75% Patient records match the incoming Patient record are returned, and 75% can be a relatively low threshold for Belongingness. We would need to understand if there are Belongingness Ratios that tend higher to the 90% within the `results` for the assumption that clustering is suboptimal to be valid, I think. This would require stratification that doesn't seem feasible or too helpful right now.
4. If `belongingness_ratio=(0.75, 0.9)` and `include_multiple_matches=false` same as 2.

alhayward: Either way, I think it would be helpful to include in the metrics 1) the number of results above the upper threshold, and 2) the number of results above the lower threshold. The statistical interpretations will vary, depending on `belongingness_ratio` and `include_multuple_matches`, as discussed above. But I think we'd just need to distinguish what these metrics are more clearly than `len(results)` and `len(scores)`. Maybe `matches_above_lower_bound` and `matches_above_upper_bound`?

ericbuckley: Yep, I've reimplemented how these are calculated.  Take a look at the new `result_counts` variable.

src/recordlinker/linking/link.py
```diff
@@ -111,38 +124,62 @@ def link_record_against_mpi(
...
-    if scores:
-        # Find the person with the highest matching score
-        matched_person, _ = max(scores.items(), key=lambda i: i[1])
-
-    sorted_scores: list[LinkResult] = [LinkResult(k, v) for k, v in sorted(scores.items(), reverse=True, key=lambda item: item[1])]
-    if not scores:
+    results: list[LinkResult] = [
+        LinkResult(k, v) for k, v in sorted(scores.items(), reverse=True, key=lambda i: i[1])
+    ]
+    result_counts["above_lower_bound"] = len(results)
```
alhayward, line 151: I believe this will only return a value > 1 if `include_multiple_matches=true`. Is that designed as intended?

src/recordlinker/linking/link.py
```diff
@@ -111,38 +124,62 @@ def link_record_against_mpi(
...
             session, record, matched_person, record.external_id, external_person_id, commit=False
         )
 
+    LOGGER.info(
+        "link results",
+        extra={
+            "person.reference_id": matched_person and matched_person.reference_id,
+            "patient.reference_id": patient.reference_id,
+            "result.prediction": prediction,
+            "result.count_patients_compared": result_counts["patients_compared"],
```
alhayward, line 178: Should we also log the number of Person clusters compared? That would put into context the number of Patients compared.

src/recordlinker/linking/link.py
```diff
@@ -55,15 +57,20 @@ def compare(
         # Evaluate the comparison function and append the result to the list
         result: float = e.func(record, patient, feature, **kwargs)  # type: ignore
         results.append(result)
-    return matching_rule(results, **kwargs)  # type: ignore
+        details[f"evaluator.{e.feature}.result"] = result
+    is_match = matching_rule(results, **kwargs)
+    details["rule.results"] = is_match
+    # TODO: this may add a lot of noise, consider moving to debug
+    LOGGER.info("patient comparison", extra=details)
```
alhayward, line 64: What is being logged here?

alhayward: Ah, I think I see - the result of the match rule (true or false for match or no match, respectively). I was thinking more of the log odds score, similarity score, log odds ratio and fuzzy threshold as helpful metrics to log in regards to feature comparison here, to make more explainable why a prediction was made. What do you think?

Also, what do you think about logging the match rule being used alongside this match rule return value? That way, there's more context to interpret the true/false result (e.g., on what matching criteria this result was generated).

alhayward (commented):
Left a few questions. Thanks for your detailed responses in the discussions!

src/recordlinker/linking/link.py
ericbuckley: This should return everything above the lower bounds, we don't take the `include_multiple_matches` variable into consideration until further down on line 165.

src/recordlinker/routes/link_router.py
```diff
@@ -20,119 +21,49 @@
...
-    if input.algorithm:
-        algorithm = algorithm_service.get_algorithm(db_session, input.algorithm)
-    else:
-        algorithm = algorithm_service.default_algorithm(db_session)
-
+    algorithm = (
+        algorithm_service.get_algorithm(db_session, label)
+        if label
+        else algorithm_service.default_algorithm(db_session)
+    )
```
cbrinson-rise8, line 32: cool way to do an inline `if` statement!

cbrinson-rise8 (approved):
Looks good to me!

PR #146, opened 2024-12-02
ID Triplet Feature

cbrinson-rise8:
## Description
- Implemenet the ID Triplet as described in [RFC-001](https://github.com/CDCgov/RecordLinker/pull/125)
- Replace MRN, SSN and DRIVERS_LICENSE with the new IDENTIFIER triplet concept introduced in RFC-001.

## Related Issues
closes #125 

## Additional Notes
- [x] Update the `/link` endpoint to accept an identifier triplet
- [x] Create blocking values on `IDENTIFIER` values
- [x] Feature match on `IDENTIFIER` and/or `IDENTIFIER:XXX` values
- [x] Add new test cases for blocking and feature matching on the new values
- [x] Update documentation in references.md regarding the new blocking key and feature

## Checklist

## Checklist for Reviewers


docs/site/reference.md
```diff
@@ -83,9 +83,13 @@ linkage evaluation phase. The following features are supported:
 
 :   The patient's email address.
 
-`DRIVERS_LICENSE`
+`IDENTIFIER`
 
-:   The patient's driver's license number.
+:   The patient's identifier. This is a catch-all for all identifiers for the patient
```
ericbuckley, line 88:
```suggestion
:   An identifier for the patient.  Matching on this will check if any identifier type/authority/value combination matches.
```

docs/site/reference.md
```diff
@@ -83,9 +83,13 @@ linkage evaluation phase. The following features are supported:
...
+
+`IDENTIFIER:<type>`
+
+:   The patient's specific identifier type. For example, `IDENTIFIER:MRN` would be the patient's medical record number.
```
ericbuckley, line 92:
```suggestion
:   The patient's specific identifier type. For example, `IDENTIFIER:MR` would be the patient's medical record number.  Unlike `IDENTIFIER`, this will ONLY compare values of a specific type.  Valid type codes can be found here http://hl7.org/fhir/R4/v2/0203/index.html.
```

src/recordlinker/assets/initial_algorithms.json
```diff
@@ -106,7 +106,7 @@
                         "CITY": 2.438553006137189,
                         "FIRST_NAME": 6.849475906891162,
                         "LAST_NAME": 6.350720397426025,
-                        "MRN": 0.3051262572525359,
+                        "IDENTIFIER": 0.3051262572525359,
```
ericbuckley, line 109:
```suggestion
                        "IDENTIFIER:MR": 0.3051262572525359,
```

src/recordlinker/assets/initial_algorithms.json
```diff
@@ -148,7 +148,7 @@
...
```
ericbuckley, line 151:
```suggestion
                        "IDENTIFIER:MR": 0.3051262572525359,
```

src/recordlinker/hl7/fhir.py
```diff
@@ -33,26 +33,19 @@ def fhir_record_to_pii_record(fhir_record: dict) -> schemas.PIIRecord:
...
-                    "value": license_number,
-                    "authority": authority
-                }
+        for code in identifier.get("type", {}).get("coding", []):
+            val["identifiers"].append({
+                "value": identifier.get("value"),
+                "type": code.get("code"),
+                "authority": identifier.get("assigner", {}).get("identifier", {}).get("value", ""),
+            })
+            break   # Sholdn't be more than 1 code
```
ericbuckley, line 48: why wouldn't there be more than 1?

src/recordlinker/schemas/identifier.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Parse type string into an IdentifierType enum
+        """
+        if value:   
+            return IdentifierType(value)
+        return value
+
+    #TODO: should we even keep this in? Can't return none for value so what to return if bad formatted SSN?
+    @pydantic.field_validator("value", mode="before")
+    def parse_value(cls, value, values):
```
ericbuckley, line 178: What is the "values" parameter?  Did you find some pydantic documentation about this?

src/recordlinker/schemas/pii.py
```diff
@@ -25,21 +26,57 @@ class Feature(enum.Enum):
...
+        # Split the feature string on ":"
+        parts = feature_string.split(":", 1)
+
+        if len(parts) == 1:
+            # If no suffix is provided, set only the attribute
+            attribute = FeatureAttribute(parts[0])
+            return cls(attribute=attribute)
+
+        # If a suffix is provided, parse both parts
+        attribute = FeatureAttribute(parts[0])
```
ericbuckley, line 73: What happens if the user specifies "FIRST_NAME:DL"?

src/recordlinker/schemas/pii.py
```diff
@@ -209,7 +232,8 @@ def model_construct(
         obj.address = [Address.model_construct(**a) for a in values.get("address", [])]
         obj.name = [Name.model_construct(**n) for n in values.get("name", [])]
         obj.telecom = [Telecom.model_construct(**t) for t in values.get("telecom", [])]
-        obj.drivers_license = DriversLicense.model_construct(**values.get("drivers_license", {}))
+        obj.identifiers = [Identifier(**i) for i in values.get("identifiers", [])]
```
ericbuckley, line 235:
```suggestion
        obj.identifiers = [Identifier.model_construct(**i) for i in values.get("identifiers", [])]
```

src/recordlinker/schemas/pii.py
```diff
@@ -322,88 +328,90 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
...
+        elif attribute == FeatureAttribute.COUNTY:
             for address in self.address:
                 if address.county:
                     yield address.county
-        elif feature == Feature.DRIVERS_LICENSE:
-            if self.drivers_license:
-                if self.drivers_license.value and self.drivers_license.authority:
-                    yield f"{self.drivers_license.value}|{self.drivers_license.authority}"
+        elif attribute == FeatureAttribute.IDENTIFIER:
+            if suffix is None:
```
ericbuckley, line 408: You could just do this in 1 if statement if you do
```
if suffix is None or suffix == identifier.type:
```

ericbuckley (dismissed):
Good start @cbrinson-rise8

src/recordlinker/hl7/fhir.py
cbrinson-rise8: There could be more than one but we don't have a way to handle an identifier with a single value and multiple code(s) objects. So I break here assuming the first one is the most recent or accurate one. Its an assumption though

src/recordlinker/schemas/pii.py
cbrinson-rise8: I was using the model construct initially but it was not mapping the IdentifierType to an enum and was leaving it as a string.

src/recordlinker/schemas/pii.py
cbrinson-rise8: Just updated to account for this by raising a value error

src/recordlinker/schemas/identifier.py
cbrinson-rise8: https://docs.pydantic.dev/latest/concepts/validators/#validation-info:~:text=the%20field_name%20property).-,Validation%20data,-%C2%B6

cbrinson-rise8: Yeah It's used to access already validated data like other members.

src/recordlinker/hl7/fhir.py
ericbuckley: Would there be any harm in creating multiple identifiers with the same value, but a different code if more than 1 code exists?

src/recordlinker/schemas/identifier.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        Parse type string into an IdentifierType enum
+        """
+        if value: 
+            return IdentifierType(value)
+        return value
+
+    #TODO: should we even keep this in? Can't return none for value so what to return if bad formatted SSN?
+    @pydantic.field_validator("value", mode="before")
+    def parse_value(cls, value, info):
```
ericbuckley, line 178:
```suggestion
    def parse_value(cls, value: str, info: ValidatorInfo):
```

src/recordlinker/schemas/identifier.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        Parse the value string
+        """
+        identifier_type = info.data["type"]
```
ericbuckley, line 182:
```suggestion
        # NOTE: Define "type" before "value" in the field definitions to guarentee that it will be available here.
        identifier_type = info.data["type"]
```

src/recordlinker/schemas/pii.py
ericbuckley: We should be able to override `Identifier.model_construct` if its not behaving exactly as you think it should.  The advantage with this over the normal constructor is that it doesn't invoke an validation calls which makes this a faster operation.  For our use case, we are trusting the data already stored in the database, so we can skip validation.

src/recordlinker/schemas/pii.py
```diff
@@ -25,21 +26,54 @@ class Feature(enum.Enum):
...
+    The schema for a feature.
+    """
+
+    model_config = pydantic.ConfigDict(extra="allow")
+
+    suffix: typing.Optional[IdentifierType] = None
+    attribute: FeatureAttribute
+
+    @classmethod
+    def parse(cls, feature_string: str) -> "Feature":
```
ericbuckley, line 54:
```suggestion
    def parse(cls, feature_string: str) -> typing.Self:
```

src/recordlinker/schemas/pii.py
```diff
@@ -25,21 +26,54 @@ class Feature(enum.Enum):
...
-    DRIVERS_LICENSE = "DRIVERS_LICENSE"
+    IDENTIFIER = "IDENTIFIER"
 
     def __str__(self):
         """
         Return the value of the enum as a string.
         """
-        return self.value
+        return self.value    
+class Feature(pydantic.BaseModel):
```
ericbuckley, line 43: Chaning Feature from an enum to a pydantic.BaseModel means we can't auto list out all the choices in the API docs anymore for features.  Since we already have an enum for the FeatureAttribtues and another for Identifier, did you consider just combining those to create a new Feature enum?  For example

```python
def all_features() -> typing.Iterator[str]:
    """
    Return a list of all possible features that can be used for comparison.
    """
    for feature in FeatureAttribute:
        yield str(feature)
        if feature == FeatureAttribute.IDENTIFIER:
            for identifier in IdentifierType:
                yield f"{feature}:{identifier}"
Feature = enum.Enum("Feature", [(f, f) for f in all_features()])
```

src/recordlinker/hl7/fhir.py
cbrinson-rise8: I don't see why not. I'll change it to behave this way instead

src/recordlinker/schemas/pii.py
cbrinson-rise8: Added in an enum for this

docs/site/reference.md
```diff
@@ -83,9 +83,13 @@ linkage evaluation phase. The following features are supported:
 
 :   The patient's email address.
 
-`DRIVERS_LICENSE`
```
ericbuckley, line 86: MRN and SSN on lines 18 and 22 need to be removed as well

src/recordlinker/schemas/pii.py
ericbuckley: Let me share a slightly different implementation.   I'm thinking we don't need both Feature and FeatureEnum, we can just use one class for both functionality and docs.  I'm not married to this implementation as it requires us to "parse" everytime we call suffix() or attribute(), so it might be better to split the suffix and attribute in the constructor so we don't have to parse everytime the functions are invoked.  However this should give you a better idea as how we can meet both needs with one class.

```python
class Feature(enum.Enum):
    """
    Enum for the different Patient attributes that can be used for comparison.
    """

    # Dynamically populate the enum members
    @classmethod
    def populate(cls):
        """Populate the Feature enum dynamically."""
        features = {}
        for f in FeatureAttribute:
            features[f.value] = f.value
            if f == FeatureAttribute.IDENTIFIER:
                for i in IdentifierType:
                    features[f"{f.value}:{i.value}"] = f"{f.value}:{i.value}"
        # Add members to the class
        cls._member_map_.update(features)
        cls._value2member_map_.update({v: cls(k, v) for k, v in features.items()})

    def attribute(self) -> FeatureAttribute:
        """Get the attribute of the feature."""
        return FeatureAttribute(self.value.split(":")[0])

    def suffix(self) -> typing.Optional[IdentifierType]:
        """Get the suffix of the feature if it is an identifier, otherwise None."""
        if self.attribute() == FeatureAttribute.IDENTIFIER:
            return IdentifierType(self.value.split(":")[1])
        return None

    def __str__(self):
        """Return the value of the enum as a string."""
        return self.value
Feature.populate()
```

src/recordlinker/schemas/algorithm.py
```diff
@@ -23,10 +23,9 @@ class Evaluator(pydantic.BaseModel):
 
     model_config = pydantic.ConfigDict(from_attributes=True, use_enum_values=True)
 
-    feature: Feature
+    feature: str = pydantic.Field(json_schema_extra={"enum": Feature.all_options()})
```
cbrinson-rise8, line 26: @ericbuckley I changed this type back to a `str` so that our tests will still pass and so that we don't have to change the way we specify the feature in the algorithm pass. I left the `pydantic.Field(...)` so all the documentation is showing up fine.

ericbuckley: The change makes sense.  Do we have any tests that show how validation works here?  For example if a user tries to create an Algorithm with a Feature of "IDENTIFIER:X" or "INVALID", what does this BaseModel do during validation?

tests/unit/schemas/test_algorithm.py
```diff
@@ -33,12 +33,13 @@ def test_validate_evaluators(self):
         evaluators = [
             {"feature": "name", "func": "func:recordlinker.linking.matchers.compare_match_any"}
         ]
-        with pytest.raises(pydantic.ValidationError):
-            AlgorithmPass(
-                blocking_keys=[],
-                evaluators=evaluators,
-                rule="func:recordlinker.linking.matchers.rule_match",
-            )
+        #TODO: changed evaluators type to a string
```
ericbuckley, line 36: [Similar to my earlier comment](https://github.com/CDCgov/RecordLinker/pull/146/files#r1906005266), it looks like we still need to address these tests.

cbrinson-rise8: good catch. I added these tests back in

src/recordlinker/schemas/algorithm.py
cbrinson-rise8: I added a validation check that throws a value error if an invalid feature is used for an evaluator. Also, updated the `test_algorithm.py` to check a few new conditions that should trigger the value error

src/recordlinker/schemas/pii.py
```diff
@@ -418,23 +432,27 @@ def blocking_keys(self, key: models.BlockingKey) -> set[str]:
 
         if key == models.BlockingKey.BIRTHDATE:
             # NOTE: we could optimize here and remove the dashes from the date
-            vals.update(self.feature_iter(Feature.BIRTHDATE))
-        elif key == models.BlockingKey.MRN:
-            vals.update({x[-4:] for x in self.feature_iter(Feature.MRN)})
+            vals.update(self.feature_iter(Feature(attribute=FeatureAttribute.BIRTHDATE)))
+        elif key == models.BlockingKey.IDENTIFIER:
+            vals.update({
+                f"{type_part}:{authority_part}:{value_part[-4:]}"
```
ericbuckley, line 438:
```suggestion
                f"{type_part}:{authority_part[:2]}:{value_part[-4:]}"
```

https://github.com/CDCgov/RecordLinker/pull/125/files#diff-565877605f59af85f1d668e71eecf89e9a43942b2d31bea13adb6ac9f30aeb43R154-R158

src/recordlinker/schemas/pii.py
```diff
@@ -418,23 +432,27 @@ def blocking_keys(self, key: models.BlockingKey) -> set[str]:
 
         if key == models.BlockingKey.BIRTHDATE:
             # NOTE: we could optimize here and remove the dashes from the date
-            vals.update(self.feature_iter(Feature.BIRTHDATE))
-        elif key == models.BlockingKey.MRN:
-            vals.update({x[-4:] for x in self.feature_iter(Feature.MRN)})
+            vals.update(self.feature_iter(Feature(attribute=FeatureAttribute.BIRTHDATE)))
+        elif key == models.BlockingKey.IDENTIFIER:
+            vals.update({
+                f"{type_part}:{authority_part[:2]}:{value_part[-4:]}"
```
ericbuckley, line 438: @cbrinson-rise8 I see that commit 1796e966b didn't break any test cases.  Would you mind adding a couple of tests to show that the IDENTIFIER blocking key is getting created correctly?  I'd like to make sure our build process fails if there is an accidental change to this logic.

cbrinson-rise8: Added a new test that should cover this.

src/recordlinker/linking/link.py
```diff
@@ -51,7 +51,7 @@ def compare(
     details: dict[str, typing.Any] = {"patient.reference_id": str(patient.reference_id)}
     for e in evals:
         # TODO: can we do this check earlier?
-        feature = getattr(schemas.Feature, e.feature, None)
+        feature = schemas.Feature.parse(e.feature)
```
ericbuckley, line 54: calling the parse method here means we have to evaluate the string every time we run a feature comparison, which can add up.  If we do it in say the `bound_evaluators` call, we only have to do it once per request.  I know that simple change would create a model->schema dependency which isn't ideal, but can you think of another way to do this earlier in the stack?

src/recordlinker/schemas/identifier.py
```diff
@@ -0,0 +1,206 @@
...
+        cls, _fields_set: set[str] | None = None, **values: typing.Any
+    ) -> typing.Self:
+        """
+        Construct a new instance of the Identifier model
+        """
+        values["type"] = IdentifierType(values["type"])
+        return super().model_construct(_fields_set=_fields_set, **values)
+
+    @pydantic.field_validator("type", mode="before")
+    def parse_type(cls, value):
```
ericbuckley, line 178: Is this method necessary?  I believe pydantic will automatically convert this into an IdentifierType because that class is an enum.   I ran the test cases with this method deleted and they all passed.  Curious as to what issues you may have saw without it?

cbrinson-rise8: Good catch. I meant to remove that. With the addition of the `model_construct` function above there's no need for this `parse_type` function anymore. Just removed it

tests/unit/routes/test_seed_router.py
```diff
@@ -38,7 +38,8 @@ def test_large_batch(self, client):
         assert sum(len(p["patients"]) for p in persons) == 1285
         assert client.session.query(models.Person).count() == 100
         assert client.session.query(models.Patient).count() == 1285
-        assert client.session.query(models.BlockingValue).count() == 10280
+        # TODO: skip this check for now seed_test.json.gz contains "mrn" and "ssn" fields. Need ot switch these to the new "identifier" format
+        # assert client.session.query(models.BlockingValue).count() == 10280
```
ericbuckley, line 42: Can you enable this one, we probably just need to increase the value to account for all the ids

tests/unit/routes/test_seed_router.py
```diff
@@ -38,7 +38,7 @@ def test_large_batch(self, client):
         assert sum(len(p["patients"]) for p in persons) == 1285
         assert client.session.query(models.Person).count() == 100
         assert client.session.query(models.Patient).count() == 1285
-        assert client.session.query(models.BlockingValue).count() == 10280
+        assert client.session.query(models.BlockingValue).count() == 8995
```
ericbuckley, line 41: This is interesting, I didn't think this was going to decrease, I thought if anything there would be more Blocking Values from the identifier change.  Any thoughts as to why this went down?

ericbuckley: Yeah something weird is going on here, when I run this query it gives me 0, meaning there are no BlockingValues in the database for the key IDENTIFIER.

```python
client.session.query(models.BlockingValue).filter(models.BlockingValue.blockingkey == 10).count()
```

cbrinson-rise8: Yeah that is weird. I'll take a look at this.

src/recordlinker/schemas/pii.py
```diff
@@ -418,23 +432,27 @@ def blocking_keys(self, key: models.BlockingKey) -> set[str]:
...
+                for x in self.feature_iter(Feature(attribute=FeatureAttribute.IDENTIFIER))
```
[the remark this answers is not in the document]
ericbuckley: Ah thats right, the default suffix value is None, so it just returns everything.  makes sense.  Sorry about that, thanks for clarifying.

ericbuckley (approved):
Thanks for all your work on this @cbrinson-rise8, I think its ready now!

Issue #139, opened 2024-11-21
rfc-002 cluster membership using log-odds scores


ericbuckley:
## Description
Create an RFC that recommends updating the belongingness ratio calculation for determining match with median log odds scores.  This change request should go into detail on why this is a better metric for determining cluster membership, what new "ratio" value users would be configuring in their business rules and how exactly this would be calculated when users select evaluation functions that don't use log-odds.

## Background / Context
#18

Issue #18, opened 2024-09-13
Algorithm changes to support more diverse set of responses


ericbuckley:
## Summary
Research into the current algorithm and how we can adjust it to return a response that is non-binary (match vs non-match).  Customers have asked for changes so they can specify a fuzzy area in the matching logic, in which it might be a match and needs to be manually reviewed.

## Acceptance Criteria
- [x] A document that describes how we propose to change the algorithm to meet the above requirements.

## Details / Tasks
There are quite a few places in the current algorithm that we could adjust to return the concept of "might be a match" to the user.  For instance:
 - If only some passes results in a match, and not all
 - If the cluster ratio is in a range
 - If the matching rule is in a range
 - If only some of the evaluators return a positive result, but not all
 - By specifying some of our thresholds as ranges

Because there are so many places in the algorithm that can be adjusted to indicated that we're not positive its a match, but it might be.  We need to make a recommendation based on where the best place(s) are to do so, and make it clear through the definition of the algorithm configuration as to where that adjustment should be made.  Alternatively, it might be possible to just return some sort of confidence interval to the call on how likely we think this is a match, and the customer can make the decision whether a manual review should take place.  Lots of options, but we should do some research and determine where the best place to make the algorithm changes are.

## Background / Context
- [Current record linkage algorithm / glossary](https://docs.google.com/document/d/1YR68Gn6WwLZ_IumsZth_tWO7B1ddGBPmcrtqnFGDzoY/edit#heading=h.w91qlqf974ic)
- [Product requirements for record linkage](https://docs.google.com/document/d/1XS5BOrV45AruNSmXbO2OqsQvK5dTLsmG1ydCkr7hTKs/edit#heading=h.s0i4xvtg1x5p)
- [Record linkage analysis](https://docs.google.com/document/d/1Gwu8V6d2V0IlcK9KQf5kIHPcgoQudln3Y330cqfwJOc/edit#heading=h.5i32wk9voy4)

alhayward:
# Short-Term

## Approach #1
[...]

PR #164, opened 2025-01-02
docs(link): rfc-002

alhayward:
## Description
* Request for Change (RFC) for Possible Match with Log Odds Score Threshold Range

## Related Issues
#139 

## Additional Notes
#18 

The language around the specific metric calculated from the distribution of Log Odds Scores (e.g., median) is intentionally broad, in case the Data Scientist on the DIBBs 4 Team would like to further experiment with other metrics beyond median.

## Checklist

## Checklist for Reviewers

### DIFF
--- a/rfc/002-log-odds-score-threshold.md
+++ b/rfc/002-log-odds-score-threshold.md
@@ -1,0 +1,115 @@
+# RFC: Possible Match with Log Odds Score Threshold Range
+
+- **Status**: Under Review
+- **Author(s)**: @alhayward
+- **Creation Date**: 2024-12-26
+- **Last Updated**: 2025-01-02
+- **RFC ID**: RFC-002
+
+---
+
+## Summary
+
+The purpose of this Request for Change (RFC) is to propose an approach for determining a Possible Match in the Record Linkage algorithm by evaluating a user-configurable threshold range for the (median) [Log Odds Score](https://cdc.sharepoint.com/:w:/r/teams/OPHDST-IRD-DIBBs/_layouts/15/doc2.aspx?sourcedoc=%7BCD3D39E1-FC95-4555-8841-15AE92ED317C%7D&file=DIBBs%20Record%20Linkage%20Algorithm%20and%20Terminology.docx&action=default&mobileredirect=true) between an incoming Patient record and the existing Patient records in a Person Cluster. This RFC proposes an alternative approach from the current implementaton, which is a user-configurable threshold range applied to the Belongingness Ratio. As such, the changes proposed in this RFC would fundamentally shift the concept of "Belongingness" in the Record Linkage algorithm by changing in the calculation that determines Cluster Membership.
+
+On the IDWA Record Linker Project in collaboration with NBS, this is Skylight's recommended Long-Term Approach for implementing Possible Match. It more robustly determines whether an incoming Patient record "belongs" to a given Person Cluster by evaluating a more granular similarity function: one of Log Odds Score. Compared to the current Short-Term Approach, this method retains more information about how similar Patient records are when comparing one incoming Patient record against *n* Patient records of an existing Person Cluster, and uses that information to make a more informed decision regarding Cluster Membership.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


rfc/002-log-odds-score-threshold.md
```diff
@@ -0,0 +1,115 @@
...
+
+* **Explainability.** While using a metric of Log Odds Score may be more understandble among certain users such as public health epidimiologists, this approach to determine Cluster Membership may be less explainable than the Belongingness Ratio (e.g., percentage of existing Patient records in the Person Cluster with which the incoming Patient record matches). However, if a STLT would like to use a more explainable approach to Record Linkage, they have the option of using the DIBBs Basic (deterministic) Algorithm to determine matches, which does not use Log Odds Scores.
+
+* **DIBBs Basic Algorithm.** If a STLT does not wish to use functions with Log Odds Scores, i.e. use the DIBBs Basic Algorithm instead, then a Match Rule function that is not evaluating the sum of Log Odds Scores against a cutoff (e.g., `recordlinker.linking.matchers.rule_probabilistic_match`) will be used (e.g., `recordlinker.linking.matchers.rule_match`). In that case, the same behavior from the Feature Comparison and Match Rule functions of the deterministic DIBBs Basic Algorithm applies. See **Caveats** in [#18](https://github.com/CDCgov/RecordLinker/issues/18) for more information.
```
ericbuckley, line 90: I'm not sure I follow this, are you suggesting we propoate the summation of feature matches and use that in our latter comparisions when a non-probablistic algorithm is configured OR are you suggesting that we remove the ability to match patient records using non-probablistic algorithms?

rfc/002-log-odds-score-threshold.md
```diff
@@ -0,0 +1,115 @@
...
+* Implementing Possible Match (and thus Manual Review) at a *pass level* can create additional manual work. This manual work may also block the Record Linkage algorithm from determining a match decision, which can create bottlenecks for required Manual Reviews as the MPI grows. To arrive at a match decision, the user may need to reconcile many Possible Matches between different permutations of record pairs.
```
ericbuckley, line 81: I'm a bit unclear as to what the proposal is here for passes.  Are you proposing we only have 1 evaluation pass per algorithm going forward?

rfc/002-log-odds-score-threshold.md
alhayward: Not quite. When a non-probabilistic algorithm is chosen (a.k.a. the DIBBs Basic Algorithm is used), I think we should not make any changes. That is, we should keep the current implementation of Possible Match for the DIBBs Basic Algorithm (Short-Term Approach with Belongingness Ratio), or any algorithms that do not use log odds scores in the feature comparison functions. This is because the Log Odds Score Threshold Range approach proposed in this RFC cannot apply when log odds scores are not used in the algorithm.

ericbuckley: So the resulting payload would change (sometimes belongingness ratio and other times log-odds distance) depending on the configuration of the Algorithm.  Speciffically the log-odds distance value would only ever be returned if all of the rule functions, in all of the passes, use a probablistic match function?

alhayward: This sounds right. If any probabilistic functions are used, log odds scores would be evaluated/returned as the "score". If no probabilistic functions are used, Belongingness Ratios would be evaluated/returned as the "score". The resulting payload can still return a general value called "score", but the value would be different depending on whether the user chose a deterministic or probabilistic algorithm.

In this way, we could return a generic "score" in the payload, regardless of whether a deterministic or probabilistic algorithm is used. Just that when the former is used, that score is the Belongingness Ratio (and should be interpreted as such), and when the latter is used, that score is the (median) Log Odds Score. In this approach, in the algorithm configuration, we could generalize the language to have users set a "score threshold" (which under the hood is either a Belongingness Ratio threshold range, or a Log Odds Score threshold range, depending on whether they chose a deterministic or probabilistic algorithm, respectively).

I think the case where the user wants to use a mix of deterministic and probabilistic match functions (e.g., deterministic in Pass 1 and probabilistic in Pass 2) is less likely and should be less supported because it's less explainable. There would be two concepts of "belongingness" (Pass 1 would use Belongingness Ratio, whereas Pass 2 would use Log Odds Score). Also, I recall we pulled Belongingness Ratio out to the algorithm configuration-level (rather than the pass-level), which would support this choice.

ericbuckley: Thanks @alhayward.  I've been thinking a lot about the score concept and believe some generic score value (different from belongingness ratio and/or log odds score) would benefit our users.  Particularly with how this could align with the concept of [similarity scores in FHIRs patient matching specification](https://www.hl7.org/fhir/us/identity-matching/patient-matching.html#scoring-matches--responders-system-match-output-quality-score).

rfc/002-log-odds-score-threshold.md
```diff
@@ -0,0 +1,115 @@
+# RFC: Possible Match with Log Odds Score Threshold Range
+
+- **Status**: Under Review
```
ericbuckley, line 3:
```suggestion
- **Status**: Deferred
```

@alhayward I'm good with this proposal as is, thanks for taking the time to write it.  However, I'm not quite sure it's ready for implementation.  I'm a little wary of a scoring system that changes based on the user business rules set in place, I think this could be confusing to our users, and would like to revisit this in the future to adjust or reaffirm that it won't cause confusion.  For the time being, I think we can merge as long as we set the status appropriately.

alhayward: Sounds good. I think that makes sense!

I'd also be curious about Brandon's thoughts on this proposal on the DIBBs 4 Team. In one of our previous DIBBs <> IDWA syncs on Record Linkage, he had a few other ideas on what this \*metric\* calculated from the Log Odds Score distribution could be (beyond median).

ericbuckley (approved):

PR #172, opened 2025-01-16
person cluster patient assignment endpoints 

### LINKED ISSUE #158: Rework patient assignment API endpoints [MPIAPI]
## Summary
Copy the existing two assign patient API endpoints to use new urls on the /person namespace

## Details / Tasks
`POST /patient/<patient_reference_id>/person` -> `POST /person`
Copy the functionality of assigning a patient (or list of patients) to a new person cluster using `POST /person`, this endpoint should accept a list of patient reference ids (eg `'{"patients": ["<patient-reference-id>"]}'`)

`PATCH /patient/<patient_reference_id>/person` -> `PATCH /person/<person-reference-id>`
Copy the functionality of assigning a patient (or list of patients) to an existing person cluster using `PATCH /person/<person-reference-id>`, this endpoint should accept a list of patient reference ids (eg `'{"patients": ["<patient-reference-id>"]}'`)

In the API docs, mark the existing `/patient/X/person` endpoints as "DEPRECATED".

NOTE: At this time, do not delete the existing /patient endpoints that we are copying, although they will be deleted in the future.

## Related Issues/PRs
#155 

## Notes / Comments
This work will complete use cases 1, 2, 3 and 6 specified in #155

ericbuckley:
## Description
Two new person API endpoints for assigning patients to different person clusters.

## Related Issues
closes #158

## Additional Notes
- Deprecating existing create_person and update_person routes in the patient router
- Created create_person and update_person in the person router to replace ones in patient router
- Updated two mpi_service functions to accept a list of patients rather than just 1
- regrouping all patient, person and seeding api endpoints into "mpi" tag in api docs

## Checklist

## Checklist for Reviewers


src/recordlinker/database/mpi_service.py
```diff
@@ -190,14 +190,19 @@ def insert_blocking_values(
         session.commit()
 
 
-def get_patient_by_reference_id(
-    session: orm.Session, reference_id: uuid.UUID
-) -> models.Patient | None:
+def get_patients_by_reference_ids(
+    session: orm.Session, *reference_ids: uuid.UUID
+) -> list[models.Patient | None]:
```
bamader, line 195: Nothing wrong in the code here, just curiosity: what's the impetus for changing this from a singular patient operating function to a plural patients? Just thinking about it pedagogically in the abstract, a single patient makes sense to me because we'll be operating on single incoming records, but I think this is an area where missing context is probably in play.

src/recordlinker/database/mpi_service.py
```diff
@@ -212,19 +217,21 @@ def get_person_by_reference_id(
 
 def update_person_cluster(
     session: orm.Session,
-    patient: models.Patient,
+    patients: typing.Sequence[models.Patient],
```
bamader, line 220: Same question here as above, when would we be bulk updating a person cluster than adding one patient at a time? This is particularly true because that could actually change performance; e.g., adding one patient, then one patient, then one patient leaves room for the linking algorithm to actually return different match results for each subsequent addition, vs bulk adding three new patients all in one cluster update runs the risk of not properly using those second and third patients to get the most accurate linkage results. It's possible one or more might be found to be non-matches for the newly updated cluster, after adding the first patient in.

src/recordlinker/routes/patient_router.py
```diff
@@ -43,24 +48,29 @@ def create_person(
...
 ) -> schemas.PatientPersonRef:
     """
+    **NOTE**: This endpoint is deprecated. Use the PATCH `/person/{person_reference_id}` endpoint instead.
+
+    **NOTE**: This endpoint will be removed in v25.2.0.
+
     Update the Person linked on the Patient.
     """
-    patient = service.get_patient_by_reference_id(session, patient_reference_id)
+    patient = service.get_patients_by_reference_ids(session, patient_reference_id)[0]
```
bamader, line 65: If we're just pulling one patient out of the returned bunch, that would seem to support having a function that does a single patient operation as the correct signature, right?

src/recordlinker/routes/patient_router.py
```diff
@@ -76,9 +86,9 @@ def delete_patient(
     """
     Delete a Patient from the mpi database.
     """
-    patient = service.get_patient_by_reference_id(session, patient_reference_id)
+    patient = service.get_patients_by_reference_ids(session, patient_reference_id)[0]
```
bamader, line 89: ditto

bamader (commented):
The code changes themselves all look fine to me. I'm wondering what the rationale is for making the functions that previously operated on a single patient or person now perform bulk retrievals / updates. That seems to actually be a fairly dangerous process (see my example above about how that could corrupt linkage results and degrade future cluster quality) if we're performing this on any kind of updates to person clusters in between processing new in coming records. This seems like a case where I'm missing some context though, such as maybe this is only used for seeding.

src/recordlinker/database/mpi_service.py
ericbuckley: The new endpoints referenced in the issue are accepting a list of patient ids.  This is an optimization as we can just get them all at once rather than querying the database N times for each patient id in the list.

https://github.com/CDCgov/RecordLinker/issues/158

src/recordlinker/database/mpi_service.py
ericbuckley: The use case we discussed with NBS is splitting a person cluster.   When NBS seeds the data, the clusters will be formed using the existing linkage decisions NBS made.  NBS asked for the ability to retroactively change the cluster if need be.  Note that all of these patient and person endpoints we are creating are purely CRUD operations NBS wanted support for so they can adjust the MPI.  None of these new endpoints will run the linkage algorithm.

I agree that allowing users to manually change the clusters could be dangerous.  On the other hand, if NBS does a good job building tools for users to make these changes, giving users the power to manually adjust could improve functionality over time.

src/recordlinker/routes/patient_router.py
ericbuckley: Yes, however this method is deprecated and will be removed in a future version.  Which reminds me I need to bump that value to v25.3.0.  I didn't bother to leave the old get_patients_by_reference_id function, because we're just going to remove this route in a couple of weeks.

src/recordlinker/routes/patient_router.py
ericbuckley: So this method will remain, so we could have a `get_patient_by_reference_id` and a `get_patient_by_reference_ids` functions to support the different user cases.  My thought was that since they are so similar, we really only need the latter.  Do you find the get element by index 0 to be confusing or do you think there is a potential for a bug in how I used it?

src/recordlinker/database/mpi_service.py
bamader: Ah I see. If they won't actually run linkage then this is probably just something that helps meet user requirements. As long as we're not overriding real linkage decisions or making choices that could value corrupt in the long term. I'm not sure whether giving users the power to manually adjust will be good or bad long term, but as long as we stick to our decision options when we do run the system I think that's what matters.

src/recordlinker/routes/patient_router.py
bamader: No I think only having one and then getting element 0 is fine. No confusion there, and we've used that pattern in the past. Mostly I think my concern was around bulk bypassing an actual linkage decision which we cleared up in the conversation above. I just wanted to make sure that when we actually do inserts into clusters after running linkage, they'll happen one at a time rather than inserting everyone at once (outside of seeding). Does that concern make sense? The issue would be:

If I'm bulk adding records A, B, C, and D all at once when running linkage, then none of those records will actually check candidate match status with the others. This could impact final clustering performance if e.g. record D was a John Smith that was lower quality than the existing person cluster for John Smith, whereas record B was a higher quality John Smith. Bulk inserting them all as part of linkage would land records B and D in the John Smith cluster, which as now reduced that person's quality. But going sequentially would put record B in John Smith first, which might then be enough to keep the lower quality record D out (which would be desirable).

bamader: To clarify, your comment above made it sound like this scenario would never arise in practice because adding more than one record to a cluster would only happen as part of seeding, which doesn't actually use the linkage algorithm. Every other time, we're going to insert sequentially and run linkage. Is that right?

bamader (approved):

src/recordlinker/database/mpi_service.py
ericbuckley: Yes it is a user (NBS) requirement, and yes they could shoot themselves in the foot here and override results.  In some of the earlier discussions we had there was talk about just manually updating the database FKs when case surveillance users make edits.  That gave me a lot of pause, as there is so many ways they can put the system in a bad state doing that.  This is a lot of power, but its power they requested, so why not give it to them in a controlled way so at least we have some safeguard around it.

My gut feeling makes me want to put some of these MPI API endpoints behind a feature flag as we thing about evolving this service into a product.  Maybe by default these `/patient` and `/person` endpoints are disabled so users don't override decisions the algorithm is making, but power users will have the ability to turn the feature flag on and gain greater control.

src/recordlinker/routes/patient_router.py
ericbuckley: None of our linkage function calls are designed to process records in parallel, nor are the API endpoints asynchronous at the moment.  However, this is hard to prevent in practice because the user could install this in a variety of different ways.

For instance, lets say the NBS foundations team decides to run the app as `uvicorn recordlinker.main:app --workers 4`, well now by default there are 4 different versions of the app running all ready to receive linkage results.  If the client starts blasting requests at the API, the API will most certainly be processing 4 in parallel!

Another option for the NBS foundations team is to scale the number of replicas of RecordLinker to a number greater than 1.  Very similar situation to the scenario above, in that the API can now handle multiple requests in parallel.

These scaling options are very common for IT and devop teams, and I'm going to assume that most will want to run parallel versions of the API for load and redundancy reasons.  All this said, it's going to come down to how the client issues calls to the API.  Are they synchronously sending records to be linked, or are they blasting them through as fast as possible?  Outside of documentation, there may be little we can control here.

PR #183, opened 2025-01-28
Update README with note about running on Windows

m-goggins:
## Description
This PR adds a brief note about running the bootstrap script on a Windows machine. Due to how WSL can be configured on a Windows machine, users may wish to ignore the directory changing OR configure the repo in the Ubuntu directory of their machine in order for the script to run.

## Related Issues
n/a

## Additional Notes
RecordLinker had not been tested on a Windows machine

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ the virtual environment to stay active in your shell.
 source scripts/bootstrap.sh
 ```
 
-Note: If you are running in WSL on a Windows machine, will need to run the bootstrap file directly with `.\scripts\bootstrap.sh` and then activate the virtual environment by running `.venv\bin\Activate.ps1`.
+Note: If you are running in WSL on a Windows machine, will need to run the bootstrap file directly with `./scripts/bootstrap.sh` and then activate the virtual environment by running `.venv/bin/activate`.
 
 ### Running the API
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 [ADD TABLE OF CONTENTS HERE]
 
-**General disclaimer** This repository was created for use by CDC programs to collaborate on public health related projects in support of the [CDC mission](https://www.cdc.gov/about/organization/mission.htm).  GitHub is not hosted by the CDC, but is a third party website used by CDC and its partners to share information and collaborate on software. CDC use of GitHub does not imply an endorsement of any one particular service, product, or enterprise. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,13 +22,17 @@ The RecordLinker is a service that links records from two datasets based on a se
 Set up a Python virtual environment and install the required development dependencies:
 NOTE: Sourcing the script is recommended over simply executing the script. This allows
 the virtual environment to stay active in your shell.
+
 ```bash
 source scripts/bootstrap.sh
 ```
 
+Note: If you are running WSL on a Windows machine, you may need to comment out line 8 in `bootstrap.sh` or configure your clone of the RecordLinker repo in the Ubuntu directory of your machine before running.
```
ericbuckley, line 30: @m-goggins thanks for adding this information.  Out of curiosity, what was the error?  Line 8 in the bootstrap.sh script is used in many of the scripts to make sure the user is in the root directory before executing the logic.  Is the `/` causing issues, or maybe `dirname`?  Wondering if we could just make a small tweak to the scripts to make them more reliable on WSL.  For instance, would this fix it?

```shell
cd "$(dirname "$(realpath "$0")")/.." || exit 1
```

m-goggins: The original error was `realpath: invalid option -- 'b'` and then as a result `-bash: .venv/bin/activate: No such file or directory` because line 8 bumped me to my `Repo` folder and not `Repo/RecordLinker` where I believe it was intended to go. The same issue persists with the code you suggested.

I wonder if this code might work to ensure a user is always at the top level of the directory for the repo?

`cd "$(git rev-parse --show-toplevel)" || exit 1`

m-goggins: Updated docs based on conversation with Eric yesterday. I was also able to get the local server running so I should be all set to start working with the code next week!

README.md
```diff
@@ -22,13 +22,17 @@ The RecordLinker is a service that links records from two datasets based on a se
 Set up a Python virtual environment and install the required development dependencies:
 NOTE: Sourcing the script is recommended over simply executing the script. This allows
 the virtual environment to stay active in your shell.
+
 ```bash
 source scripts/bootstrap.sh
 ```
 
+Note: If you are running in WSL on a Windows machine, will need to run the bootstrap file directly with `.\scripts\bootstrap.sh` and then activate the virtual environment by running `.venv\bin\Activate.ps1`.
```
ericbuckley, line 30: @m-goggins maybe I'm getting my terminology mixed up, but did we determine its `.venv/bin/activate` on WSL and `.venv\bin\Activate.psq` on PowerShell?  Also note the slashes, I thought in WSL you use the standard Linux forward slash.

m-goggins: Oh yep, you're correct! I had both terminals opened and copied from the wrong one.

ericbuckley (approved):
Thanks for making our documentation better @m-goggins!

Issue #184, opened 2025-01-29
Build msodbc drivers into ARM64 images


ericbuckley:
## Summary
msodbc18 can now be installed on ARM64 (eg Mac M1) platforms, we should remove the if block in the Dockerfile to exclude this process for those platforms.

## Acceptance Criteria
- [ ] Remove the if platform == arm64 check in the Dockerfile

PR #190, opened 2025-02-03
Build msodbc drivers into arm64 images

m-goggins:
## Description
msodbc18 can now be installed on ARM64 (eg Mac M1) platforms so we can remove the if block in the Dockerfile to exclude this process for those platforms.

## Related Issues
#184 

## Additional Notes
I'm not sure if we have a good way to test the success of this change but if I followed the Teams conversation correctly, we have confirmed that NBS was able to get this working.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,12 +27,12 @@ RUN pip install --upgrade pip
 # Conditionally install ODBC driver for SQL Server.
 # There is no ODBC driver for linux/arm64 architecture, so SQL Server support
 # is limited to linux/amd64 architecture
-RUN if [ "$USE_MSSQL" = "true" ] && [ "$(dpkg --print-architecture)" = "amd64" ]; then \
-        apt-get install -y gnupg2 apt-transport-https && \
-        curl https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /etc/apt/trusted.gpg.d/microsoft.gpg && \
-        curl https://packages.microsoft.com/config/debian/11/prod.list | tee /etc/apt/sources.list.d/mssql-release.list && \
-        apt-get update && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -46,8 +44,8 @@ RUN pip install "$(printf '%s' ".[${ENVIRONMENT}]")"
 
 # Conditionally install OpenTelemetry packages if USE_OTEL is true
 RUN if [ "$USE_OTEL" = "true" ]; then \
-        pip install opentelemetry-distro opentelemetry-exporter-otlp && \
-        opentelemetry-bootstrap -a install; \
+    pip install opentelemetry-distro opentelemetry-exporter-otlp && \
```
ericbuckley, line 47: Any reason for removing the indentation? I know most shells don't care about indentation in blocks, but from a readability perspective I think it helps.

m-goggins: hmm I didn't make any intentional changes; looks like this was done automatically with the Docker formatter. Are you using another formatter?

m-goggins: It looks like mine was just the default Docker formatter that comes with VS Code. I was able to turn it off for dockerfiles and add the indentations back.

ericbuckley: I'm actually intrigued by the idea of linting on Docker files and shell scripts as well (we have a few of both of those).  I don't want to hold up this change, but if you have any information on what formatter VS Code uses under the hood let me know.

ericbuckley (approved):

Dockerfile
m-goggins: I looked into it and there is no default Docker formatter/linter; when I saved, the default whitespace and tabspace settings in VS Code were being applied. That said, there are some formatters available for Docker. I haven't done a comprehensive look but [hadolint](https://github.com/hadolint/hadolint) look promising for Dockerfiles.

Issue #155, opened 2024-12-16
Design MPI API


ericbuckley:
## Summary
Add and amend the existing MPI API endpoints to wholistically address a wide variety of MPI management use cases.

## Acceptance Criteria
- [ ] A list of use cases for when and why an upstream user would want to adjust the MPI
- [ ] A proposal for a new set of API endpoints to manage the MPI

## Background / Context
Record Linker was initially designed to address the automatic loading and organizing patient documents from upstream providers (hospitals, clinics, labs, etc).  While this is a very important use case to cover, its not the only source of information that an MPI should be able to track.  There are many use cases in which the downstream users, case investigators and epidemiologists, want to manually adjust the MPI based on information they've acquired post document ingestion.

A [few additional endpoints](https://cdcgov.github.io/RecordLinker/api-docs.html#tag/patient) have been added to help address some of these use cases, however they don't address all the use cases.  Let's revisit these, and the use cases that are necessary for upstream users, to create a new, complete set of API endpoints to manage the MPI data.

ericbuckley:
# Proposal for MPI management API endpoints

The following API changes are being proposed to give users full control over making changes to their MPI. 
Note, these endpoints will replace the [3 existing API endpoints](https://cdcgov.github.io/RecordLinker/api-docs.html#tag/patient) under `/patient`.

## API Endpoints

* **A1**: `POST /person` Create a new person cluster
* **A2**: `GET /person/orphaned` List orphaned persons
* **A3**: `GET /person/<person-reference-id>` List all patients on a person cluster
* **A4**: `POST /person/<person-reference-id>` Add a new patient to a person cluster
* **A5**: `PATCH /person/<person-reference-id>` Assign an existing patient to a person cluster
* **A6**: `DELETE /person/<person-reference-id>` Delete a person cluster
* **A7**: `POST /person/<person-reference-id>/merge` Merge person clusters
* **B1**: `GET /patient/orphaned` List orphaned patients
* **B2**: `GET /patient/<patient-reference-id>` View record details of a patient
* **B3**: `PUT /patient/<patient-reference-id>` Update record details of a patient
* **B4**: `DELETE /patient/<patient-reference-id>` Delete a patient

## Use Cases

### Use Case 1
After a possible match, a user has decided to assign a patient to a new person.
```
(A1) POST /person -d '{"patients": ["<patient-reference-id>"]}'
```

### Use Case 2
After possible match, a user decides to assign a patient to an existing person.
```
[...]

PR #203, opened 2025-02-07
Add API endpoint for merging person clusters

### LINKED ISSUE #161: API endpoint for merging of person clusters [MPIAPI]
## Summary
Add a new API endpoint to allow users to merge person clusters.

## Details / Tasks
`POST /person/<person-reference-id>/merge -d '{"persons": ["<person-reference-id1>", "<person-reference-id2>"]}'`
Create the new API endpoint to merge existing person records (could be a list of them) into the referenced one.

## Related Issues/PRs
#155

## Notes / Comments
This will complete use cases 4 and 5 in #155

m-goggins:
## Description
This PR adds a new endpoint to merge person clusters as such `POST /person/<person-reference-id>/merge -d '{"persons": ["<person-reference-id1>", "<person-reference-id2>"]}'`. 

Two questions that I want to raise about the PR:

1. After the person clusters are merged, e.g., `person-reference-id1` and `person-reference-id2` into `person-reference-id` as above, `person-reference-id1` and `person-reference-id2` are not currently cleaned up (deleted from the database). Originally I wasn't sure if this should occur as a separate step, but after working on the endpoint and our discussion about additional manual work yesterday, I think I should add a cleanup step after merging to remove `person-reference-id1` and `person-reference-id2`. In theory, this cleanup could occur on the routine cleanup that we've talked about but not yet implemented, but I like the "cleanup as you go if you can" approach for now. Thoughts?

2. Does anyone have strong opinions on creating similar functions that have different inputs and returns like `get_person_by_reference_id` and `get_persons_by_reference_ids`? They are used for slightly different purposes but I could condense into a single function if folks feel it improves clarity. 

## Related Issues
Fixes #161


ericbuckley:
@m-goggins re question 1; this is accounted for in [use case 12 of the MPI API design](https://github.com/CDCgov/RecordLinker/issues/155).  My original thought was, let's make this an explicit DELETE rather than an implicit one.  It's true that these old person clusters are orphaned and no longer in use, however we can not definitely say that the user's systems have purged all references to these clusters.  By making it an explicit step to DELETE the orphaned person clusters, we have more trust that the user is doing the same with their references.

This isn't a hard requirement, nor am I married to that implementation.  It may turn out that automatic cleanup is easier, or we could just provide a flag in the payload to indicate that automatic removal is preferred.

src/recordlinker/routes/person_router.py
```diff
@@ -75,3 +100,58 @@ def update_person(
...
+
+    if person is None:
+        raise fastapi.HTTPException(status_code=fastapi.status.HTTP_404_NOT_FOUND)
+
+    # Get all persons by person_reference_id that will be merged
+    persons = persons_by_reference_ids_or_422(session, data.person_reference_ids)
+    person_ids = [person.id for person in persons]
+
+    # Get all of the patients from the person clusters to be merged
+    patients = patients_by_person_id_or_422(session, person_ids)
```
ericbuckley, line 150: In what scenario would we raise a 422 on this line?  If I'm reading this correctly, any invalid person_reference_id would be caught on line 146 and result in a raised exception there, so I think the list of person_ids should be safe by the time we get to line 150.

src/recordlinker/routes/person_router.py
```diff
@@ -75,3 +100,58 @@ def update_person(
...
+
+    # Merge the person clusters
+    person = service.update_person_cluster(session, patients, person, commit=False)
```
ericbuckley, line 153: If you wanted to do this in one query rather than 2 (merge lines 150 and 153), I think you could write a mpi_service function to do that.
```
UPDATE patients p
SET p.person_id = X
WHERE p.person_id in (Y, Z ,...)
```

src/recordlinker/routes/person_router.py
m-goggins: Hmm I guess it could depend on how we do clean up of the person clusters. I guess I was thinking of the scenario where patients have been (re) assigned a new person cluster but their empty person cluster is still around...does that make sense? I might not have enough understanding of the current development so I'm open to removing this check if you're comfortable with it.

src/recordlinker/routes/person_router.py
m-goggins: I like that and will update accordingly!

src/recordlinker/routes/person_router.py
ericbuckley: In that case, let's say where a user is trying to merge a person cluster with 0 patients.  I would think we get the empty list back.  Are you thinking that in the case where a user tries to merge cluster B into cluster A, but cluster B is empty that we should return a 422 rather than a noop?

src/recordlinker/database/mpi_service.py
```diff
@@ -292,6 +328,27 @@ def update_person_cluster(
...
+    session: orm.Session,
+    person: models.Person,
+    person_ids: typing.Sequence[int],
+    commit: bool = True,
+) -> models.Person:
+    """
+    Update the person_id for all Patients associated with a Person.
+    """
+    # Check that the person_ids are valid integers
+    if not all(isinstance(pid, int) for pid in person_ids):
```
ericbuckley, line 341: There's nothing wrong with this check, but it doesn't match the style of the rest of the codebase currently, and I'm fairly confident that line 342 could never be executed.

Right now, we have a [CI check that runs mypy](https://github.com/CDCgov/RecordLinker/blob/main/.github/workflows/check_types.yml) on all the types.  Given that person_ids is defined as a sequence of ints on line 334, mypy will fail at any point in the stack if it finds a value that doesn't adhere to that.  Mypy can't be exhaustive of course, since python isn't a statically typed language, and all it takes is one function somewhere in the stack with no types on its signatures.  That said, our codebase has so far thoroughly defined types on pretty much every method (I want to say 100%, but I'm sure there is an exception) and to my knowledge SQLAlchemy has also defined types for all its methods.

Do you happen to know of any situations, other than what I just described, in which mypy wouldn't catch this?

m-goggins: I also thought that mypy would catch this type of error but when I added this [test](https://github.com/CDCgov/RecordLinker/pull/203/files#diff-10fac7af215509d82caf0a8313dcb74b43f9e0693d1204a6a7f793fa136eff0dR824-R826), I realized that it doesn't. I could switch the type to `list[int]` if you think that's a better approach.

ericbuckley: [By default, mypy is only analyzing the source code](https://github.com/CDCgov/RecordLinker/blob/main/pyproject.toml#L115-L117).  We could switch mypy over to analyze both, but I have a feeling that lots of issues might arise.

For me, its not about which one is better, I think they both add value.  It's about style.  My concern is that if we start mixing and matching (sometimes using type hints and other times writing parameter checks), it will become confusing to the team down the road which one to use and when.

As a team, we never formalized around using python type hints, it just organically grew.  Maybe the codebase is of a certain size now where we need to be explicit about that.  Do you have any thoughts around python type checks vs parameter checks.  Do you prefer one over the other, or maybe have a preference for specific situations?

m-goggins: That's a good point and helpful context to have. I agree that mixing and matching would be confusing to developers in the future. 

My preference would be to keep the type checks but perhaps add some parameter checks for the function parameters to ensure early failures, particularly if we can do it before accessing the DB. For example, checking that the `data` parameter for `merge_person_clusters` is not empty, etc. We ran into several issues once we started working in LAC's environment because we did not have adequate existence or empty list checks (for example), which is informing my thinking.

That said, my preference above does not apply to line 341 😆 so I'm happy to remove it and appreciate the conversation helping me solidify my opinion.

rfc/001-identifier-triplets.md
```diff
@@ -199,3 +199,8 @@ field has slightly different behavior, but does allow for maximum flexibility.
 - Variations in authority values could lead to false negatives in comparisons (eg `CA` vs `California`).
 - Variations in value formats could lead to false negatives in comparisons (eg `123-45-6789` vs `123456789`).
 - Blocking values can't be specified per type, so if we want to block on identifiers we need to block on all.
+
+## Updates
```
ericbuckley, line 203: small thing, but mind updating the date on line 6

src/recordlinker/database/mpi_service.py
```diff
@@ -263,16 +264,51 @@ def get_patients_by_reference_ids(
     return [patients_by_id.get(ref_id) for ref_id in reference_ids]
 
 
+def get_patients_by_person_ids(
```
ericbuckley, line 267: Is this function still needed?

src/recordlinker/database/mpi_service.py
```diff
@@ -263,16 +264,51 @@ def get_patients_by_reference_ids(
...
+def get_persons_by_reference_ids(
+    session: orm.Session, person_reference_ids: typing.Sequence[uuid.UUID]
+) -> typing.Sequence[models.Person] | None:
+    """
+    Retrieve multiple Persons by their person reference IDs.
+    """
+    if not person_reference_ids:
+        return None
+    query = select(models.Person).where(models.Person.reference_id.in_(person_reference_ids))
+    return session.scalars(query).all()
```
ericbuckley, line 309: Let's say we call this method with two person_reference_ids, a valid one (meaning this UUID matches some person cluster in the DB) and an invalid one (its a UUID, but doesn't match any person cluster in the DB).

In that situation, you will get back a list of person objects, its just the list will be of length 1 not 2.  I believe that's going to cause problems for the router call, because its only raising a 422 if the value is None.

m-goggins: I hadn't thought of that scenario. I added back in a 422 check for getting the person_reference_ids which should raise a 422 if any of the passed person_reference_ids are invalid. It seems safer to fail in this scenario, though I'm curious if NBS has provided any insight about how they would prefer to handle that scenario; should we proceed with a merge when possible or fail if any of the person reference ids are invalid?

src/recordlinker/routes/person_router.py
```diff
@@ -31,10 +31,35 @@ def patients_by_id_or_422(
...
+    """
+    Retrieve the Patients by their reference ids or raise a 422 error response.
+    """
+    persons = service.get_persons_by_reference_ids(session, *person_reference_ids)
+    if None in persons:
+        raise fastapi.HTTPException(
+            status_code=fastapi.status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=[
+                {
+                    "loc": ["body", "patients"],
```
ericbuckley, line 56:
```suggestion
                    "loc": ["body", "person_reference_ids"],
```

src/recordlinker/database/mpi_service.py
```diff
@@ -292,6 +306,23 @@ def update_person_cluster(
...
 
 
+def update_patient_person_ids(
+    session: orm.Session,
+    person: models.Person,
+    person_ids: typing.Sequence[int],
+    commit: bool = True,
+) -> models.Person:
+    """
+    Update the person_id for all Patients associated with a Person.
```
bamader, line 316: Nit: could we update the language on this comment? The code itself seems to be updating to the ID of the provided person the person_ids of all patients currently associated to a person_id in the given list (a grotesque tongue twister, to be sure). This comment reads to me like we're going to be passing in a list of patients and then just batch setting them all to point to the given person ID. Not blocking by any means, I just think the explicit clarity would reduce the double takes.

src/recordlinker/routes/person_router.py
```diff
@@ -75,3 +100,34 @@ def update_person(
...
+)
+def merge_person_clusters(
+    person_reference_id: uuid.UUID,
+    data: typing.Annotated[schemas.PersonRefs, fastapi.Body()],
+    session: orm.Session = fastapi.Depends(get_session),
+) -> schemas.PersonRef:
+    """
+    Merges Person cluster(s) into the Person cluster referenced by `person_reference_id`.
+    """
+    # Get the person_reference_id the person clusters will be merged into
```
bamader, line 118: Nit: should be getting the Person itself, since we pass in the person_reference_id

src/recordlinker/routes/person_router.py
```diff
@@ -75,3 +100,34 @@ def update_person(
...
+    per = service.get_person_by_reference_id(session, person_reference_id)
+
+    if per is None:
+        raise fastapi.HTTPException(status_code=fastapi.status.HTTP_404_NOT_FOUND)
+
+    # Get all persons by person_reference_id that will be merged
```
bamader, line 124: Nit: since we're using person_reference_id as a more general construct here to refer to "an ID that uniquely references a given person cluster", what would you say to changing the parameter name from `person_reference_id` to something like `merge_into_id`? Right now, person_reference_id is getting a little overloaded to me.

bamader (commented):
Mostly looks solid to me, just some nits on language and comment cleanups that I think could help reduce the cognitive burden of reading some of this stuff

ericbuckley:
> @m-goggins re question 1; this is accounted for in [use case 12 of the MPI API design](https://github.com/CDCgov/RecordLinker/issues/155). My original thought was, let's make this an explicit DELETE rather than an implicit one. It's true that these old person clusters are orphaned and no longer in use, however we can not definitely say that the user's systems have purged all references to these clusters. By making it an explicit step to DELETE the orphaned person clusters, we have more trust that the user is doing the same with their references.
> 
> This isn't a hard requirement, nor am I married to that implementation. It may turn out that automatic cleanup is easier, or we could just provide a flag in the payload to indicate that automatic removal is preferred.

@m-goggins PR is looking great, thanks for all your work on this so far.  I wanted to revisit the above with regard to auto removal of the person cluster when its empty.  What do you think about updating the payload to accept a flag for this action?  Do you think that would be helpful for users?

src/recordlinker/database/mpi_service.py
m-goggins: This is ugly. What about "Update the person_id for all Patients by current person_id."?

src/recordlinker/routes/person_router.py
m-goggins: Yes, I like that suggestion.

m-goggins:
>  What do you think about updating the payload to accept a flag for this action? Do you think that would be helpful for users?

@ericbuckley That sounds like a good approach to me. I can make that change.

src/recordlinker/database/mpi_service.py
bamader: Yeah that's better (mine was hideous I know 😅 ; was just trying to capture the actual logic of what was happening. I know the prepositional ordering of "updating _to the ID of the provided person_ the person_ids..." is grammatically correct but it always feels so wrong to write it out).

bamader (dismissed):
Approving so I don't block, since we've got agreement on new comments.

tests/unit/routes/test_person_router.py
```diff
@@ -112,3 +117,74 @@ def test_with_patients(self, client):
...
+        assert response.json()["person_reference_id"] == str(person1.reference_id)
+
+    def testInvalidPersonIdType(self, client):
+        response = client.post(
+            "/person/123/merge/",
+            json={"person_reference_ids": [str(uuid.uuid4())]},
+        )
+        assert response.status_code == 422
+
+    def test1InvalidPersonId(self, client):
```
ericbuckley, line 148: Could we add a test case to show that the person ids specified in the body of the payload can't be the same as the one specified in the URL?

tests/unit/routes/test_person_router.py
```diff
@@ -112,3 +117,85 @@ def test_with_patients(self, client):
...
+
+        response = client.post(
+            f"/person/{person1.reference_id}/merge/",
+            json={"person_reference_ids": [str(person2.reference_id), str(uuid.uuid4())]},
+        )
+        assert response.status_code == 422
+
+    def testPersonIdsNottheSame(self, client):
+        # Test that a 422 will be raised if the merge_into_id is in the person_reference_ids
+        person1 = models.Person()
```
ericbuckley, line 167:
```suggestion
        person1 = models.Person(reference_id=uuid.uuid4())
```
By default, the reference_id is `null`.  If we persist that, SQLAlchemy will [create a default value](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/models/mpi.py#L22) for us.  Alternatively, we can just initialize the object with a value and skip the `session.add`, `session.flush` operations.

ericbuckley (approved):
This looks great @m-goggins thanks for all your work on this!

Issue #201, opened 2025-02-07
compare_probabilistic_exact_match


ericbuckley:
## Summary
Create a new evaluator function named `compare_probabilistic_exact_match`, this should behave just like the fuzzy_match variant but return a 0 if its not an exact match and the entire log odds score if it is.  This will be useful for enum fields like SEX and RACE.

## Acceptance Criteria
- [ ] A new matching function
- [ ] Registration in the FeatureFunc class
- [ ] New test cases in test_matchers

PR #209, opened 2025-02-10
Create function for probabilistic exact match

bamader:
## Description
This PR adds a matcher function for evaluating a probabilistic exact match. For a given feature field in two patient records, the matcher awards maximum log odds points for that field if the two fields are exactly the same and 0 points otherwise.

## Related Issues
#201

### DIFF
--- a/docs/site/reference.md
+++ b/docs/site/reference.md
@@ -174,7 +174,7 @@ existing Patient with the GIVEN_NAME of ["John", "D"].
     0.0. This is useful when performing probabilistic comparisons (which score a possible match's strength by
     accumulating a sum of link weights) on fields for which fuzzy similarity doesn't make sense, such as fields
     defined by an enum (e.g. Sex). Use the kwargs parameter to specify the log-odds ratios based on training.
-    Example: `{"kwargs": {"thresholds": {"FIRST_NAME": 0.8}, "log_odds": {"FIRST_NAME": 6.8}}}`
+    Example: `{"kwargs": {"log_odds": {"SEX": 6.8}}}`
 
 `func:recordlinker.linking.matchers.compare_probabilistic_fuzzy_match`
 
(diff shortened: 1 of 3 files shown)


docs/site/reference.md
```diff
@@ -166,6 +166,16 @@ existing Patient with the GIVEN_NAME of ["John", "D"].
...
 
+`func:recordlinker.linking.matchers.compare_probabilistic_exact_match`
+
+:   Determines if a Feature Field has the same value in two different patient records. If the two fields agree
+    exactly (i.e. are exactly the same), then the function returns the full extent of the log-odds weights for 
+    the particular field with which it was called. If the two fields do not exactly agree, the function returns
+    0.0. This is useful when performing probabilistic comparisons (which score a possible match's strength by
+    accumulating a sum of link weights) on fields for which fuzzy similarity doesn't make sense, such as fields
+    defined by an enum (e.g. Sex). Use the kwargs parameter to specify the log-odds ratios based on training.
+    Example: `{"kwargs": {"thresholds": {"FIRST_NAME": 0.8}, "log_odds": {"FIRST_NAME": 6.8}}}`
```
ericbuckley, line 177:
```suggestion
    Example: `{"kwargs": {"thresholds": {"SEX": 0.8}, "log_odds": {"SEX": 6.8}}}`
```
Just thinking it might be better to show an example here with an enum field rather than a free text field to hammer home when we think its useful.

ericbuckley: Actually, is a threshold parameter necessary for this function?

bamader: You're 100% right, no threshold needed! That's what I get for trying to bounce back and forth between branches 😅

src/recordlinker/linking/matchers.py
```diff
@@ -189,11 +192,47 @@ def compare_fuzzy_match(
...
+                break
+    return agree * log_odds
+
+
 def compare_probabilistic_fuzzy_match(
     record: PIIRecord, patient: Patient, key: Feature, **kwargs: typing.Any
 ) -> float:
     """
-    ...
+    Compare the same Feature Field in two patient records, one incoming and one
```
ericbuckley, line 230: Thanks for adding this!

ericbuckley (approved):

PR #211, opened 2025-02-10
Reverse identifiers for comparisons

### LINKED ISSUE #200: reverse Identifier for comparisons
## Summary
When using jaro-winkler to compare strings, more weight is put into the details at the front of the string rather than at the back.  With Identifier, the current string format is "type:authority:value".  We should put more weight into the value, as to not inflate scores that are matching identifiers with just the same type.

## Acceptance Criteria
- [ ] Reverse the Identifier string when yielding from feature_iter

## Details / Tasks
The new format of identifiers should be "{value}:{authority}:{type}"

m-goggins:
## Description

Since Jaro-Winkler is the default similarity measure for string comparisons, more weight is given to differences at the start of a string. As such, this PR modifies the way that we yield and store Identifiers for comparison. Previously, the Identifier format was `type:authority:value` but now it is `value:authority:type` to give more weight to the value. For example, when comparing two MRNs with values, `1234` and `2345`, we will now give more weight to `1234` vs `2345` than the type, e.g., `MR` vs `MR`. This should yield better differentiation between strings.

## Related Issues
Fixes #200 

## Additional Notes
I updated the language in the RFC to reflect the new format and added a note at the bottom to indicate that the RFC has been changed.

### DIFF
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -39,8 +39,9 @@ class FeatureAttribute(enum.Enum):
         """
         Return the value of the enum as a string.
         """
-        return self.value    
-    
+        return self.value
+
+
 class Feature(pydantic.BaseModel):
     """
     The schema for a feature.
@@ -88,6 +89,7 @@ class Feature(pydantic.BaseModel):
                 for identifier in IdentifierType:
                     options.append(f"{feature}:{identifier}")
         return options
+
 
 class Sex(enum.Enum):
     """
@@ -190,6 +192,7 @@ class Telecom(pydantic.BaseModel):
             return None
         return self.value
 
+
 class PIIRecord(pydantic.BaseModel):
     """
     The schema for a PII record.
@@ -224,7 +227,7 @@ class PIIRecord(pydantic.BaseModel):
         obj.name = [Name.model_construct(**n) for n in values.get("name", [])]
         obj.telecom = [Telecom.model_construct(**t) for t in values.get("telecom", [])]
         obj.identifiers = [Identifier.model_construct(**i) for i in values.get("identifiers", [])]
-        
+
         return obj
 
     @pydantic.field_validator("external_id", mode="before")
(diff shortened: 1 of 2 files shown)


src/recordlinker/schemas/pii.py
```diff
@@ -392,25 +395,37 @@ def blocking_keys(self, key: models.BlockingKey) -> set[str]:
...
             vals.update(self.feature_iter(Feature(attribute=FeatureAttribute.BIRTHDATE)))
         elif key == models.BlockingKey.IDENTIFIER:
-            vals.update({
-                f"{type_part}:{authority_part[:2]}:{value_part[-4:]}"
-                for x in self.feature_iter(Feature(attribute=FeatureAttribute.IDENTIFIER))
-                for type_part, authority_part, value_part in [x.split(":", 2)]
-            })
+            vals.update(
+                {
+                    f"{value_part[-4:]}:{authority_part[:2]}:{type_part}"
```
ericbuckley, line 400: I think this is fine for consistency, but technically blocking keys don't have the same problem since we are also check for equality when writing our SQL where statements.

tests/unit/schemas/test_pii.py
```diff
@@ -228,50 +228,92 @@ def test_feature_iter(self):
...
+            "suffix",
+            "suffix2",
+        ]
+        assert list(record.feature_iter(pii.Feature(attribute=pii.FeatureAttribute.COUNTY))) == [
+            "county"
+        ]
+        assert list(
+            record.feature_iter(pii.Feature(attribute=pii.FeatureAttribute.IDENTIFIER))
+        ) == ["123456::MR", "123-45-6789::SS", "D1234567:VA:DL"]
+        # assert list(record.feature_iter(pii.Feature(attribute=pii.FeatureAttribute.IDENTIFIER))) == ["MR::123456", "SS::123-45-6789", "DL:VA:D1234567"]
```
ericbuckley, line 290:
```suggestion
```

ericbuckley (approved):

ericbuckley (approved):

PR #218, opened 2025-02-14
RFC-003: skip values

### LINKED ISSUE #215: RFC for skip values
## Summary
Develop an RFC that proposes a system for user-defined skip values.  The proposal should address a range of situations, starting from simple "string values to ignore" to more complex "ignore when the context indicates so".

## Acceptance Criteria
- [ ] A new RFC with a proposal for implementing user-defined skip values

## Details / Tasks
Largely speaking, there are two phases in the system in which we can achieve this.
- Data validation, the process in which incoming data is cleansed and normalized.
- Linkage matching, the process in which records are compared to one another.
It should be strongly considered that the latter is preferable from a flexibility perspective.  As it gives our users options to adjust the values over time. 

[Common skip values to consider](https://cdc.sharepoint.com/:w:/r/teams/OPHDST-IRD-DIBBs/_layouts/15/Doc.aspx?sourcedoc=%7B543cf7ff-8992-41aa-8595-5b49194b5c3a%7D&action=editnew).

## Background / Context
Healthcare organizations often use placeholder values in documents to indicate missing information, such as the name “John Doe” or an address of “unknown.” When these values appear in incoming data, they should be excluded from comparisons. In other words, two patients should not be considered a match solely because both are named “John Doe.” Additionally, we want this functionality to be customizable, allowing organizations to extend or modify the default values provided by Record Linker to fit their specific needs.

ericbuckley:
## Description
A proposal to introduce system changes allowing users to define skip conditions for placeholder values to be ignored during the linkage process.

## Related Issues
closes #215

## Checklist

## Checklist for Reviewers


rfc/003-skip-values.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Unresolved Questions
+
+A key question to consider is whether the concept of skip value conditions accounts for all
+possible scenarios. Do we need more complex rules that evaluate multiple attributes
+simultaneously? For example, should an `ADDRESS` of "123 Main St" be skipped only if the
+`ZIP` is also "99999"? While supporting rules of this complexity wouldn't require
+significant changes to the specification, it would increase the implementation complexity.
+We recommend pursuing this type of **and** logic only if there is a clear and specific need
+for it.
```
m-goggins, line 147: Does NBS have skip values? I imagine theirs would give us insight into how complex the skip rules might be.

rfc/003-skip-values.md
```diff
@@ -0,0 +1,153 @@
...
+
+The second alternative is to apply specific rules during evaluation to determine whether a
+value should be used in a comparison or ignored as a placeholder. The main flaw with this
+approach is the lack of context. For example, if comparing the `FIRST_NAME` field and the
+incoming record has the value "John", this is typically a valid name for comparison.
+However, if the `LAST_NAME` is "Doe", it may be recognized as a placeholder. Performing the
+cleaning upfront allows for a comprehensive review of all data fields before evaluation,
+ensuring that additional context is considered.
+
+## Risks and Drawbacks
```
m-goggins, line 110: I think another potential risk is that it might make the results of matching harder to interpret because there's no record of the skip value field being skipped during the linking process. A user would have to know that the record they are looking at has a field in the skip values and that it wasn't used to link that record. I suppose some of this might be resolved with the log-odds scores and how we end up displaying them to users. For example, if a `JOHN DOE` comes it and we display that `FIRST_NAME` and `LAST_NAME` contribute nothing to the log-odds score, we could also indicate that it's because `NAME` contained a skip value.

That said, I still agree that not changing the data is the best approach. Just want to flag that we might need to think about how we display some of this to users down the line if we get to the point of having a UI.

rfc/003-skip-values.md
```diff
@@ -0,0 +1,153 @@
...
+The `feature` specifies the field to which the condition applies, while `value` indicates
+the case-insensitive value to skip. A wildcard value `*` can be used for feature to apply
+the condition to all fields. Example:
+
+```json
+...
+"skip_values": [
+  {
+    "feature": "NAME",
+    "value": "John Doe"
```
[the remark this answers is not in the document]
ericbuckley: That would make it easier to enter multiple conditions for the same feature, so yeah I'll make that update.  I've also been wondering if this should be an exact match, or if we should be building in some sort of fuzzy comparisons to handle misspellings.

rfc/003-skip-values.md
ericbuckley: Its a good thought, I brought it up with their team today.  I believe the answer is no, but they'll get back to us on this.

rfc/003-skip-values.md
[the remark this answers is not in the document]
m-goggins: I'd worry about similar enough (but real) names getting flagged as skip values, e.g. John Do, and losing out on valuable information for matching. I'm open to hearing other opinions but I think exact values would be safer in this instance.

rfc/003-skip-values.md
ericbuckley: Yeah its a good point, and something we should account for when designing the weighted confidence algorithm.  If we can give feedback to the users on WHY a decision was made, including skip value matches is going to be very helpful.

m-goggins (dismissed):

rfc/003-skip-values.md
```diff
@@ -0,0 +1,157 @@
...
+proposes a system that allows users to specify conditions under which these values should
+be excluded from comparisons.
+
+## Motivation
+
+Healthcare organizations often encounter incomplete or missing data and commonly use
+placeholder values to indicate this, such as "Unknown" or "Not Specified." In some cases,
+placeholders are field-specific, like using "John Doe" or "Jane Doe" for missing patient
+names. Comparing two records with the placeholder name "John Doe" does not imply they
+belong to the same individual, as the value is a stand-in rather than an actual data point.
```
bamader, line 25: You might note it later on, but there's a related issue too: comparing a John Doe to John Smith shouldn't be taken as first name agreement, either, which means we need a skip value system that's aware across fields (in an ideal world)

rfc/003-skip-values.md
```diff
@@ -0,0 +1,157 @@
...
+users later update their skip values, as any removed data would no longer be available in
+the MPI for future use.
+
+The second alternative is to apply specific rules during evaluation to determine whether a
+value should be used in a comparison or ignored as a placeholder. The main flaw with this
+approach is the lack of context. For example, if comparing the `FIRST_NAME` field and the
+incoming record has the value "John", this is typically a valid name for comparison.
+However, if the `LAST_NAME` is "Doe", it may be recognized as a placeholder. Performing the
+cleaning upfront allows for a comprehensive review of all data fields before evaluation,
+ensuring that additional context is considered.
```
bamader, line 108: I'm not sure I follow the last sentence here. I definitely get the idea that we want to apply skip values before evaluation, but where is this exactly sitting? In my mind it's the `feature_iter` yield function (i.e. for each of `FIRST_NAME` and `LAST_NAME`, if `FIRST_NAME == John and LAST_NAME == Doe yield ''`), but is that what you mean here?

rfc/003-skip-values.md
```diff
@@ -0,0 +1,157 @@
...
+certain matches were made.
+
+## Implementation Plan
+
+For the purposes of this RFC, we will not be overly prescriptive about the implementation
+details. However, the work can be broadly divided into three tasks:
+1. A new `NAME` feature will be created, that will allow us to specify skip conditions
+    for the entirety of the name specified. (This likely won't be used for evaluation,
+    as its still preferable to compare the first and last names separately, but users
+    will have that option)
```
bamader, line 135: Creating another field for name seems somewhat redundant to me, since we have first and last already, but not enough to be a blocker. I would highlight some possible edge cases, though, in which a middle name is supplied that either breaks up or otherwise modifies a skip value. For example, if "John Doe" is a skip value but a provider's system makes them enter "John A. Doe" to provide a middle initial, comparing skip fields on Name alone would now have a problem with John Does coming from places with different requirements, a problem that could be avoided if we just checked on First and Last. If we've thought this through and have ways around it, that's good with me, but just want to highlight that if we have a NAME feature that combines all parts of a name we might introduce some edge risk.

bamader (approved):
I like the approach generally but wanted to flag a possible concern around using NAME rather than jointly considering First and Last. Nothing that's blocker level, but just raising.

rfc/003-skip-values.md
ericbuckley: The new feature, `NAME` as noted below, is intended to address this.

rfc/003-skip-values.md
ericbuckley: The cleaning step as indicated in step 3 of the Implementation Plan, will likely need to duplicate some of the logic in `feature_iter`. It will accept a record object and set of cleaning rules, then return a new record object sans any features matched in the cleaning rules.

It's this new copy, returned from the cleaning function, that will be eventually passed to the `feature_iter`.

rfc/003-skip-values.md
ericbuckley: There is some redundancy across features already.  FIRST_NAME vs GIVEN_NAME.  EMAIL vs TELECOM.  IDENTIFIER vs IDENTIFIER:SS.  We've had some of these discussions with NBS, and they understood and agreed that having that flexibility was helpful.  Although we all agreed that it could add confusion, so documentation between the differences is key.  NAME vs FIRST_NAME and NAME vs LAST_NAME, would just be two more of these cases.

I don't think most users would want to use NAME in their evaluation or blocking steps, nor would we ever use it as a default.  I considered coming up with a new value set for skip_values, but thought the differences between what can be compared and what can be cleaned could make things confusing to users if they are slightly off.  However, we already make that difference between blocking keys and evaluators, so maybe a third set of values isn't that more confusing.  I guess my thought was, its easy to add a feature_iter for NAME, while its likely not useful, it doesn't hurt that much to do it.

I hadn't thought of the "John A. Doe" use case, thanks for bringing that up.  A user could always do this if they know one provider commonly submits.

```json
"skip_values": [{"feature": "NAME", "values": ["John Doe", "John A. Doe"]}]
```

Another startegy here could be to use regexs or globs, for example:

```json
"skip_values": [{"feature": "NAME", "values": ["John % Doe"]}]
```

Thoughts?

rfc/003-skip-values.md
bamader: Okay, so this is where we'd do the thing I mentioned in my comment above (comparing John Doe to John Smith)? I.e. the cleaning function would see that one of the records is a John Doe and then "blank" out the name values, before `feature_iter` ever gets ahold of it?

bamader: If that's the case, then will we need to apply the cleaning function to blocked candidates returned from the MPI? In other words, if we've previously processed some John Does, and they get created as new Person Clusters in the MPI, then anytime in the future we get a record who might block into John Doe, we'll need to make sure we don't do a comparison on the fetched candidates. Although, writing that out, there might be some trickier rules around blocking with John Doe anyway.

ericbuckley: You know, my original thought was we only need to "clean" the incoming record.  As long as we clean one of the two objects being compared, we'd be safe.  But if the incoming record is "John Smith" and the blocked candidate is "John Doe" then we'd get a perfect match on FIRST_NAME, when we shouldn't!

You're 💯 right, we need to clean both the incoming record AND the blocked candidates.

rfc/003-skip-values.md
ericbuckley: Yeah, I was thinking the [glob syntax](https://docs.python.org/3/library/glob.html) would suffice.

bamader (approved):
Looks good, I think we're there with this!

ericbuckley:
Moving this to ON HOLD for the time being.

Issue #163, opened 2024-12-20
Create API endpoints to manage orphaned data [MPIAPI]


ericbuckley:
## Summary
Create 3 new API endpoints to assign with managing orphaned (unlinked) data.

## Details / Tasks
`DELETE /person/<person-reference-id>`
This should delete an EMPTY person cluster, if the person cluster has members this should result in an error.

`GET /person/orphaned`
A paginated list of all the persons with no members.

`GET /patient/orphaned`
A paginated list of all the patients not yet assigned to a person cluster.

## Background / Context
`GET /person/orphaned` could be an expensive query to run, give this query a lot of consideration so we know it will scale with large databases. 

## Related Issues/PRs
#155

## Notes / Comments
This will complete use cases 12 and 13 in #155.

PR #221, opened 2025-02-19
Add endpoint for deleting orphaned persons

m-goggins:
## Description
This PR adds an API endpoint for deleting persons who have no associated patients. It works by checking that the provided person_reference_id is valid and then checking if the person has any associated patients. Persons with associated patients are not deleted. 

## Related Issues
Addresses 1 of the 3 API endpoints listed in #163

## Additional Notes
The other API endpoints in #163 will be in follow-up PRs.

### DIFF
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -370,7 +370,7 @@ def delete_persons(
         session.commit()
 
 
-def check_person_for_patients(session: orm.Session, person: models.Person) -> bool:
+def check_person_for_patients(session: orm.Session, person: models.Person) -> int | None:
     """
     Check if a Person has at least 1 associated Patient.
     """
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -370,9 +370,9 @@ def delete_persons(
         session.commit()
 
 
-def check_person_for_patients(session: orm.Session, person: models.Person) -> int | None:
+def check_person_for_patients(session: orm.Session, person: models.Person) -> bool:
     """
     Check if a Person has at least 1 associated Patient.
     """
     query = select(literal(1)).filter(models.Patient.person_id == person.id).limit(1)
-    return session.execute(query).scalar()
+    return True if session.execute(query).scalar() is not None else False
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -9,6 +9,7 @@ import typing
 import uuid
 
 from sqlalchemy import insert
+from sqlalchemy import literal
 from sqlalchemy import orm
 from sqlalchemy import select
 from sqlalchemy.sql import expression
@@ -367,3 +368,11 @@ def delete_persons(
         session.delete(person)
     if commit:
         session.commit()
+
+
+def check_person_for_patients(session: orm.Session, person: models.Person) -> bool:
+    """
+    Check if a Person has at least 1 associated Patient.
+    """
+    query = select(literal(1)).filter(models.Patient.person_id == person.id).limit(1)
+    return session.execute(query).scalar()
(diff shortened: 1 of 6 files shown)


src/recordlinker/routes/person_router.py
```diff
@@ -170,3 +170,39 @@ def merge_person_clusters(
...
+    # Check that person_reference_id is valid
+    person = service.get_person_by_reference_id(session, person_reference_id)
+
+    if person is None:
+        raise fastapi.HTTPException(status_code=fastapi.status.HTTP_404_NOT_FOUND)
+
+    # Check if the person has associated patients
+    if person.patients:
+        raise fastapi.HTTPException(
+            status_code=fastapi.status.HTTP_403_FORBIDDEN,
```
ericbuckley, line 196: Thoughts on using a 422 here instead of a 403.  I think redoc automatically creates documentation for 422 when we use pydantic for validation.

ericbuckley: I checked, and it just shows the 204 and 422.  We can also update the docs and add a 3rd response type.

Screenshot 2025-02-20 at 16 23 33

src/recordlinker/routes/person_router.py
```diff
@@ -170,3 +170,39 @@ def merge_person_clusters(
...
+    Delete an empty Person from the MPI database.
+    """
+    # Check that person_reference_id is valid
+    person = service.get_person_by_reference_id(session, person_reference_id)
+
+    if person is None:
+        raise fastapi.HTTPException(status_code=fastapi.status.HTTP_404_NOT_FOUND)
+
+    # Check if the person has associated patients
+    if person.patients:
```
ericbuckley, line 194: I'm 99% sure this is fine and is preferable because it reads well, however as is this retrieves all the linked patients out of the database.  We really only need to run
```sql
SELECT 1 
FROM mpi_patient 
WHERE person_id = X 
LIMIT 1;
```

Can you think of any situations where the person cluster would be so large, multiple thousands, where it be advantageous to just to a quick existence check rather than retrieve all?

src/recordlinker/routes/person_router.py
m-goggins: I opted to add the docs for 403 since I think that error will make more sense than getting a 422 in this instance. Also allows us more flexibility moving forward. Let me know if you think `ErrorResponse` and `ErrorDetail` should be in a different schema file (I stuck them in mpi)

src/recordlinker/routes/person_router.py
m-goggins: That's a good point! From our experience in LAC, most person clusters were n~13 so definitely not getting up there. I modified the query to match your suggestion.

src/recordlinker/database/mpi_service.py
```diff
@@ -367,3 +368,11 @@ def delete_persons(
         session.delete(person)
     if commit:
         session.commit()
+
+
+def check_person_for_patients(session: orm.Session, person: models.Person) -> int | None:
```
ericbuckley, line 373: Given the docstring, what do you think about this change?
```suggestion
def check_person_for_patients(session: orm.Session, person: models.Person) ->bool:
```

src/recordlinker/schemas/__init__.py
```diff
@@ -54,4 +56,7 @@
     "PersonCluster",
     "PersonGroup",
     "PersonRefs",
+    "ErrorRespone",
```
ericbuckley, line 59:
```suggestion
```

ericbuckley (approved):
looks good, just a couple of minor suggestions

src/recordlinker/database/mpi_service.py
m-goggins: I think that was my original intention, because you're right, it is more aligned with the docstring.

ericbuckley (approved):

PR #222, opened 2025-02-20
Add Endpoint for getting orphaned patients

m-goggins:
## Description
This PR adds an endpoint that returns a paginated list of patients not yet assigned to a person cluster. 
## Related Issues
#163 

## Additional Notes
I opted for cursor/keyset pagination to hopefully ensure that the query scales as the DB grows. 

Another thing I considered was adding another flag (or endpoint) for returning all results at once (something like `while next is not None`, but we may encounter lots of results and may want to consider some safeguards.

### DIFF
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -388,13 +388,15 @@ def get_orphaned_patients(
     cursor (in the form of a patient reference_id) is provided, only retrieve Patients
     with a reference_id greater than the cursor.
     """
-    cursor_uuid = uuid.UUID(cursor) if cursor else None
     query = (
         select(models.Patient)
         .where(models.Patient.person_id.is_(None))
-        .where(models.Patient.reference_id > cursor_uuid if cursor_uuid else True)
         .order_by(models.Patient.reference_id)
         .limit(limit)
     )
+
+    # Apply cursor if provided
+    if cursor:
+        query = query.where(models.Patient.reference_id > cursor)
 
     return session.execute(query).scalars().all()
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -367,3 +367,12 @@ def delete_persons(
         session.delete(person)
     if commit:
         session.commit()
+
+
+def get_orphaned_patients(
+    session: orm.Session,
+) -> typing.Sequence[schemas.PatientRefs]:
+    """
+    Retrieve all the orphaned Patients in the MPI database.
+    """
+    return session.scalars(select(models.Patient).where(models.Patient.person_id.is_(None))).all()
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -381,7 +381,7 @@ def check_person_for_patients(session: orm.Session, person: models.Person) -> bo
 def get_orphaned_patients(
     session: orm.Session,
     limit: int | None = 50,
-    cursor: str | None = None,
+    cursor: uuid.UUID | None = None,
 ) -> typing.Sequence[models.Patient]:
     """
     Retrieve orphaned Patients in the MPI database, up to the provided limit. If a
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -9,6 +9,7 @@ import typing
 import uuid
 
 from sqlalchemy import insert
+from sqlalchemy import literal
 from sqlalchemy import orm
 from sqlalchemy import select
 from sqlalchemy.sql import expression
@@ -367,6 +368,14 @@ def delete_persons(
         session.delete(person)
     if commit:
         session.commit()
+
+
+def check_person_for_patients(session: orm.Session, person: models.Person) -> bool:
+    """
+    Check if a Person has at least 1 associated Patient.
+    """
+    query = select(literal(1)).filter(models.Patient.person_id == person.id).limit(1)
+    return True if session.execute(query).scalar() is not None else False
 
 
 def get_orphaned_patients(
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -371,8 +371,21 @@ def delete_persons(
 
 def get_orphaned_patients(
     session: orm.Session,
+    limit: int | None = 50,
+    cursor: str | None = None,
 ) -> typing.Sequence[models.Patient]:
     """
-    Retrieve all the orphaned Patients in the MPI database.
+    Retrieve orphaned Patients in the MPI database, up to the provided limit. If a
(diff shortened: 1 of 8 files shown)


src/recordlinker/routes/patient_router.py
```diff
@@ -57,6 +57,35 @@ def create_patient(
...
+    patients = service.get_orphaned_patients(session, limit, cursor)
+    if not patients:
+        return schemas.PaginatedPatientRefs(
+            patients=[], meta=schemas.PaginatedMetaData(next_cursor=None, next=None)
+        )
+
+    # Prepare the meta data
+    next_cursor = patients[-1].reference_id if len(patients) == limit else None
+    base_url = str(request.url).split("?")[0]
+    next_url = f"{base_url}?limit={limit}&cursor={next_cursor}" if next_cursor else None
```
ericbuckley, line 81:
```suggestion
    next_url = f"{request.base_url}?limit={limit}&cursor={next_cursor}" if next_cursor else None
```
I think you can delete line 80, `base_url` is already a property on the request.

src/recordlinker/routes/patient_router.py
```diff
@@ -57,6 +57,35 @@ def create_patient(
...
+
+    return schemas.PaginatedPatientRefs(
+        patients=[p.reference_id for p in patients if p.reference_id is not None],
```
ericbuckley, line 84: In what cases would reference_id be NULL?

tests/unit/routes/test_patient_router.py
```diff
@@ -170,3 +170,73 @@ def test_get_patient(self, client):
...
+        patient1 = models.Patient(person=None, reference_id=ordered_uuids[0])
+        patient2 = models.Patient(person=None, reference_id=ordered_uuids[1])
+        patient3 = models.Patient(person=None, reference_id=ordered_uuids[2])
+        client.session.add_all([patient1, patient2, patient3])
+        client.session.flush()
+
+        # Retrieve 1 patient after patient1, return cursor for patient2
+        response = client.get(f"/patient/orphaned?limit=1&cursor={patient1.reference_id}")
+        assert response.status_code == 200
+        print()
```
ericbuckley, line 210: can we delete the debugging statements from 210 to 214

src/recordlinker/routes/patient_router.py
```diff
@@ -57,6 +57,35 @@ def create_patient(
...
+    session: orm.Session = fastapi.Depends(get_session),
+    limit: int | None = fastapi.Query(50, alias="limit", ge=1, le=1000),
+    cursor: uuid.UUID | None = fastapi.Query(None, alias="cursor"),
+) -> schemas.PaginatedPatientRefs:
+    """
+    Retrieve patient_reference_id(s) for all Patients that are not linked to a Person.
+    """
+    patients = service.get_orphaned_patients(session, limit, cursor)
+    if not patients:
+        return schemas.PaginatedPatientRefs(
```
ericbuckley, line 74: In the case that the user sends an unknown UUID (say they have a reference to an old Patient that has since been deleted).  The response will be 200 and empty, rather than a 422 indicating it's not valid.   Thoughts?

src/recordlinker/database/mpi_service.py
```diff
@@ -376,3 +376,27 @@ def check_person_for_patients(session: orm.Session, person: models.Person) -> bo
...
+    query = (
+        select(models.Patient)
+        .where(models.Patient.person_id.is_(None))
+        .order_by(models.Patient.reference_id)
+        .limit(limit)
+    )
+
+    # Apply cursor if provided
+    if cursor:
+        query = query.where(models.Patient.reference_id > cursor)
```
ericbuckley, line 400: This operation gives me pause, although I don't see anything wrong with it.

UUID1, 6 and 7 are all orderable since they are time based, but UUID4 is random. My initial thought tells me this wouldn't work, or at least wouldn't perform well at scale.

However, the only two alternatives I can think of also suffer from KNOWN performance issues.

1) Replace cursor with an offset, however offsets require paginating over the entire dataset, so they become expensive as the value gets really large.
2) Use the PK, which is an int and sortable, instead of the reference_id.  However, the user doesn't know the PK, so this would require an extra query to convert the incoming reference_id into a PK.  If we do validate the incoming reference_id, which isn't the case now, then we'd get the query for free and it perform just as well.

I don't think we need to change anything here, but I did want to flag this for discussion.

m-goggins: Yes, I definitely considered those options as well! I did not implement #1 for the same reason you state; I don't think it would hold up well at all considering that the patient table is likely to grow the quickest. 

For #2, I initially considered it because of the natural sorting but thought that the PK might not return consistent results (if say row 16 was deleted and you were expecting the 10 results after id=10, but did not get 16). However, users are sending the reference_id and not the PK, so they wouldn't be expecting any particular id(s) to be returned. In this case, I'm happy to make the change to look up the `id`, especially since I agree with your comment about validating the UUID and getting the query for free.

src/recordlinker/routes/patient_router.py
m-goggins: In that case, I'd want to return a 422. If we add a check that the cursor is a valid UUID, we can also use that query to get the `id` for use as the cursor.

src/recordlinker/database/mpi_service.py
ericbuckley: I can see issues with both creating and deleting records in between paginating results.  In general, I consider (and hopefully most API users) listing results to be a snapshot in time, and the second they are returned nothing is guaranteed to be the same.  So in the case of a `LIST, DELETE, LIST` the second list call could have different results, and the same goes for `LIST, CREATE, LIST`.

I think in both cases, using the `id ` or the `reference_id`, having a delete operation in between pagination calls can create unexpected results.  However, because the `id` value is auto-incremented.  Having a creation operation in between pagination calls should only create unexpected results when using the `reference_id` as the sort key, as we know any new patient records will get added to the end of the sort if we are sorting on `id`.

tests/unit/database/test_mpi_service.py
```diff
@@ -892,3 +893,60 @@ def test_check_person_for_patients(self, session):
...
+        patient3 = models.Patient(person=models.Person(), data={})
+        session.add_all([patient1, patient2, patient3])
+        session.flush()
+
+        assert len(mpi_service.get_orphaned_patients(session, limit=1)) == 1
+        assert len(mpi_service.get_orphaned_patients(session, limit=2)) == 2
+        assert len(mpi_service.get_orphaned_patients(session, limit=3)) == 2
+
+    def test_get_orphaned_patients_cursor(self, session: Session):
+        # ordered_uuids = [uuid.uuid4() for _ in range(4)]
```
ericbuckley, line 928:
```suggestion
```

ericbuckley (approved):
Great work @m-goggins

PR #225, opened 2025-02-25
Add endpoint to return orphaned persons

m-goggins:
## Description
The PR adds an endpoint to return a paginated list of all the persons with no members.

While working on this PR I also made some small changes to `get_orphaned_patients`:

- added an explicit `ORDER_BY` after discovering some inconsistent results when I included a `limit` but not a `cursor`. I also updated the tests accordingly.
- genericized `schemas.PaginatedRefs` so that the paginated results for both get orphaned patients and persons are the same.

## Related Issues
#163 

## Additional Notes
I spent a good amount of time thinking about how we should execute this query to make it as efficient as possible. My working assumptions were that: 
1) the patient table will be larger than the person table, 
2) both tables will be large, and 
3) orphaned persons will be relatively rare.

I considered two approaches, "LEFT JOIN" (which I ultimately landed on) and "NOT EXISTS", but I am open to hearing others.

LEFT JOIN approach:
```
SELECT p.*
FROM mpi_person p
LEFT JOIN mpi_patient pt ON pt.person_id = p.id
WHERE pt.id IS NULL
``` 

From `EXPLAIN QUERY PLAN`, we can see that we are scanning `mpi_person`, using a Bloom filter when scanning the larger patient table to more quickly eliminate rows that definitely do not match any rows in Person, and uses a covering index on `mpi_patient.id`, which should keep things as quick as possible. 

NOT EXISTS approach:
```
SELECT *
FROM mpi_person p
WHERE NOT EXISTS (
    SELECT 1 
    FROM mpi_patient pt 
    WHERE pt.person_id = p.id
)
```
This approach is less efficient because of the subquery executes multiple times checking for matching rows in Patient for each row in Person. It would be more efficient if we add an index on Patient.person_id, but I still think LEFT JOIN is a better choice given the assumptions stated above (especially #1). 

I also briefly considered a view, but given the number of updates to the Patient and Person tables, I don't think this is our best path forward. 

All that said, I am open to other approaches and would love to get folks' thoughts.

## Checklist

## Checklist for Reviewers


src/recordlinker/database/mpi_service.py
```diff
@@ -386,10 +386,39 @@ def get_orphaned_patients(
...
+    query = (
+        select(models.Person)
+        .outerjoin(models.Patient, models.Patient.person_id == models.Person.id)
+        .filter(models.Patient.id.is_(None))
+        .order_by(models.Person.id)
+    )
+    if cursor:
+        query = query.filter(models.Person.id > cursor)
+
+    query = query.limit(limit)
```
ericbuckley, line 422: Just curious, but is there a reason you did the limit after the cursor filter in this function, but before in the get_orphaned_patients?  I think it results in the same query, but wondering if there might be some subtle difference in SQLAlchemy.

ericbuckley (dismissed):
Looks great @m-goggins.  Thanks for wrapping up all the MPIAPI work!

src/recordlinker/database/mpi_service.py
m-goggins: I'll add a comment, that's a good idea.

src/recordlinker/database/mpi_service.py
```diff
@@ -386,10 +386,41 @@ def get_orphaned_patients(
...
+        .outerjoin(models.Patient, models.Patient.person_id == models.Person.id)
+        .filter(models.Patient.id.is_(None))
+        .order_by(models.Person.id)
+    )
+    if cursor:
+        query = query.filter(models.Person.id > cursor)
+
+    query = query.limit(
+        limit
+    )  # limit applied after cursor to ensure the limit is applied after the JOIN and starts from the cursor afte rthe join
```
ericbuckley, line 424:
```suggestion
    )  # limit applied after cursor to ensure the limit is applied after the JOIN and starts from the cursor after the join
```

PR #229, opened 2025-02-27
Implement RMS

### LINKED ISSUE #236: Implement Relative Match Score
## Summary
Replace the belongingness ratio score with a relative match score value that is based on log odds comparisons of different evaluation fields.

## Acceptance Criteria
- [x] Update the algorithm configuration to accept a `possible_match_window` on each pass
- [x] Remove the `beloningness_ratio` parameter on the algorithm configuration
- [x] Implement the new `link.link_record` method

~~- [ ] Merge the `link.LinkResult` and `schemas.LinkResult` class into one, keeping the latter~~ (moved to follow up)
- [x] Update the `schemas.LinkResult` class to include the `relative_match_score` value, `pass_number` ~~and `context`~~ of feature evaluations (context moved to follow up)
- [x] Remove the `schemas.Prediction` class
- [x] Create the `schemas.MatchGrade` enum
- [x] Add sufficient logging to show values calculated for the evaluation phase and when selecting a "best pass" from the cluster table.
- [x] Update documentation in site/design.md to include info on RMS and remove information about the belongingness ratio

## Details / Tasks
The details in how the new relative match score (RMS) is calculated can be found in this [document](https://cdc.sharepoint.com/:w:/t/OPHDST/EUnCSsVRlCVEkNhtIRqwq0wB2fgaigGnXLos98UfNorpgw?wdOrigin=TEAMS-MAGLEV.p2p_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1741200335478&web=1).  Use that as a guide for implementing the new `link_record` function, which will accept the same input and more or less produce the same output.

The `schemas.LinkResult` class needs to change to capture the differences between a belongingness ratio score and RMS.
1. Remove the existing `belongingness_ratio` value
2. Add the `relative_match_score` value
3. Add a pass number to indicate which pass produced the best score for the cluster (1 indexed)
~~4. Add a context list, this is a list of all the feature evaluations for the pass that generated the highest score.  Each item in the list will have the feature name and the weighted score associated to that feature.~~

~~The `MatchGrade` enum should allow for ordering, that is, I should be able to show that `MatchGrade.CERTAIN >  MatchGrate.POSSIBLE`.  Consider using the [total_ordering decorator](https://docs.python.org/3/library/functools.html#functools.total_ordering) for implementation.  For the hierarchy; `CERTAIN > POSSIBLE > CERTAINLY_NOT`.~~

When selecting the best results from the evaluation table, first compare on MatchGrade.  If two passes for the same cluster have the same MatchGrade, then select the result with the highest RMS.

## Background / Context
[RFC-002](https://github.com/CDCgov/RecordLinker/blob/main/rfc/002-log-odds-score-threshold.md)

## Testing Considerations
Please include algorithm test results of the 84 test cases NBS shared with us, using both the existing codebase and the RMS implementation.

bamader:
# Implement RMS

## Description
This PR provides all code necessary to implement a log-odds based RMS comparison to evaluate match quality. This will replace belongingness ratio as the internal decision-driver the algorithm uses to find the highest quality match.

## Related Issues
Closes #236 

## Checklist

## Checklist for Reviewers


src/recordlinker/assets/initial_algorithms.json
```diff
@@ -25,7 +25,12 @@
                         "func": "func:recordlinker.linking.matchers.compare_probabilistic_fuzzy_match"
                     }
                 ],
-                "rule": "func:recordlinker.linking.matchers.rule_probabilistic_match",
+                "rule": "func:recordlinker.linking.matchers.rule_probabilistic_sum",
+                "possible_match_window": [
+                    0.8,
+                    0.925
+                ],
+                "maximum_points": 13.2001963043,
```
ericbuckley, line 33: Does this need to be specified?  We know the log odds values and we know the evaluators in use for a particular pass, it seems like the algorithm should just be able to calculate this.

src/recordlinker/database/mpi_service.py
```diff
@@ -59,6 +62,8 @@ def get_block_data(
         )
 
     # Using the subquery of unique Patient IDs, select all the Patients
+    # NOTE: We probably apply the filter here to throw away patients with
+    # non-empty but wrong blocking fields?
```
ericbuckley, line 66: What does wrong mean in this context?  The join clause will exclude any patient record that doesn't create an exact match.

src/recordlinker/assets/initial_algorithms.json
bamader: Great catch, not in the slightest do we need to define this here. I just started in this file and then changed some structure later, so I can go back and define this calculation on the fly.

src/recordlinker/database/mpi_service.py
bamader: Ah this is something completely unrelated to log odds windows that I was just sticking in as a comment. This is building onto the point I made yesterday that because we're no longer calculating belongingness ratio we don't have to fetch all the patients associated with a Person cluster anymore, even if they didn't meet the blocking criterion. Marcelle correctly observed that if we pulled back people from the Person cluster who didn't meet blocking conditions but because they had missing values, we would solve half of the missingness problem for problem. So this comment is in service of that: it seemed like all the patients in each blocked Person got pulled back, so we'd then want to filter that to only include those patients who exact matched blocking and patients from blocked Persons who didn't block because their fields were missing.

src/recordlinker/linking/link.py
```diff
@@ -30,13 +31,76 @@
...
+        # Case 3: new grade is certain, and since grades didn't enter the 
+        # equality if, existing grade is only possible
+        elif grade == 'certain':
+            self._do_update(earned_points, rms, mmt, cmt, grade)
 
 
 def compare(
     record: schemas.PIIRecord, patient: models.Patient, algorithm_pass: models.AlgorithmPass
-) -> bool:
+) -> typing.Union[bool, float]:
```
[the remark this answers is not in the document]
bamader: I completely agree here. I think I was focusing too hard on preserving back compatibility, because we can still get "auto-matching" / non-possible-match mode by just setting the window bounds to be equal. I would propose this function to return the summation of log-odds (and maybe we rename the function `compare_and_score` to emphasize that) and then the broader link caller handle that sum. Because the comparison proper isn't actually affected by user thresholds, so it doesn't care about normalization or whatever they set.

bamader:
@ericbuckley @m-goggins  Updated code to make all tests pass (there's still one set of three tests that involved belongingness that I'm not quite sure what to do with, but will continue noodling).

I also ran the `algorithm tests` using the certain hierarchy I established, and we doubled our performance! If we were getting 33% last time (two out of the six cases), we now get 4 out of the 6 cases correct. We correctly get 3 out of the 5 match cases, and correctly flag the invalid birthdate. The cases we miss are (1) the record has first and last name switched (there's nothing we can do about that, since first and last are blocking fields in one pass [they fail exact matching] and then evaluation fields in another [but they earn 0 points]); and (2) a test case where someone named `Tho-mas` fails to match to a cluster that contains 2 identical patients except for their first names, `Thomas` and `ThoMas`. I believe if we implemented basic string normalization on incoming names (e.g. standardize casing, remove numbers and punctuation) we would catch this case. We don't even have to do it at the persistence level so that data is still preserved as supplied; we could apply it on the fly during record evaluation, just like we proposed with skip values. This would allow us to still correctly model names that might actually have punctuation (e.g. the Arabic surname `Al'Charif`, also sometimes written `Al-Charif`) but wouldn't penalize one of those representations over the other. I believe this is a bug we should fix, but wanted to drop my findings here  before I actually took off for the day.

bamader:
Updating to mark all code here complete. All tests have been adjusted to match the new algorithm and everything passes. This is, from my view, "merge ready" code.

tests/unit/linking/test_link.py
```diff
@@ -139,11 +141,12 @@ def test_compare_identifier_match(self):
...
             ],
-            rule="func:recordlinker.linking.matchers.rule_probabilistic_match",
-            kwargs={"log_odds": {"IDENTIFIER": 0.35}, "true_match_threshold": 0.3},
+            rule="func:recordlinker.linking.matchers.rule_probabilistic_sum",
+            possible_match_window=(0.8, 0.925),
+            kwargs={"log_odds": {"IDENTIFIER": 0.35}},
         )
 
-        assert link.compare(rec, pat, algorithm_pass) is True
+        assert link.compare(rec, pat, algorithm_pass) == 0.35
```
m-goggins, line 149:
```suggestion
        assert link.compare(rec, pat, algorithm_pass) == algorithm_pass.kwargs["log_odds"]["IDENTIFIER"]
```

m-goggins: nit: this makes it clearer that the resulting score is expecting to be equal to the log-odds and not normalized to 1.0. Might consider for other tests as well if you think it's a good idea.

tests/unit/linking/test_link.py
```diff
@@ -186,16 +189,17 @@ def test_compare_identifier_with_suffix(self):
...
-            rule="func:recordlinker.linking.matchers.rule_probabilistic_match",
-            kwargs={"log_odds": {"IDENTIFIER": 0.35}, "true_match_threshold": 0.3},
+            rule="func:recordlinker.linking.matchers.rule_probabilistic_sum",
+            possible_match_window=(0.8, 0.925),
+            kwargs={"log_odds": {"IDENTIFIER": 0.35}},
         )
 
         #should pass as MR is the same for both
-        assert link.compare(rec, pat, algorithm_pass) is True
+        assert link.compare(rec, pat, algorithm_pass) == 0.35
```
m-goggins, line 198: Hm. this is beyond the scope of this PR but I'm wondering if we should be treating all identifiers equally. It makes sense to me to give full points if the patient agrees with any identifiers of the same type in a record since you can have multiple MRNs, driver's licenses, etc. However, SSN should not be different because it's a national ID. I'm wondering if we should treat SSN differently.

src/recordlinker/linking/link.py
```diff
@@ -30,15 +31,79 @@
...
+        and Cluster A scored an RMS of 0.922, that would grade as 'possible'.
+        But despite the higher RMS, Cluster A actually accumualted more points
+        and stronger separation already, so we don't want to downgrade.)
+        3. If the new grade is certain but the existing grade is not, we
+        *always* update. Being a 'certain' match is a stronger statement and
+        thus more worth saving. Consider the example above with the passes
+        as normal. It would be better to save the Pass 2 RMS of 0.918 that 
+        graded as 'Certain' than it would be to keep the Pass 1 RMS of 0.922
+        that only graded 'Possible,' since the user's previous profiling found
+        these matches higher quality.
```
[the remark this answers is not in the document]
bamader: Yeah I actually have no need to keep this around at all. Back when this was intended as a branch "to-convince," this was doing some demonstration, but since we're all on board, it's now not needed I think.

src/recordlinker/linking/link.py
```diff
@@ -58,11 +123,25 @@ def compare(
...
+    
+    rule_result = matching_rule(results, **kwargs)
+    details[f"rule.{matching_rule.__name__}.results"] = rule_result
     # TODO: this may add a lot of noise, consider moving to debug
     LOGGER.info("patient comparison", extra=details)
-    return is_match
+    return rule_result
+
+
+def grade_result(rms: float, mmt: float, cmt: float) -> str:
```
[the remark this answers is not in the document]
bamader: I like `grade_rms` the most! It ties it back into the broader workflow.

src/recordlinker/linking/link.py
```diff
@@ -122,54 +206,74 @@ def link_record_against_mpi(
...
+                        # Let the dynamic programming table track its own updates
+                        scores[person].handle_update(
+                            cluster_median, rms, minimum_match_threshold, certain_match_threshold, match_grade
+                        )
+    
+    results: list[LinkResult] = sorted(scores.values(), reverse=True, key=lambda x: x.rms)
+    certain_results = [x for x in results if x.grade == 'certain']
+    # re-assign the results array since we already have the higher-priority
+    # 'certain' grades if we need them
+    results = [x for x in results if x.grade == 'possible']
```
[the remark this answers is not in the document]
bamader: In the function's final return call, we use `results` as the placeholder for all types of results to make the signature. That's where `certain_results` will get reassigned if the `if` triggers, and if it doesn't, that's where the `possible_results` already live. And of course if no results come back, the array is empty. I'm going to leave this variable as is because it keeps the return signature the cleanest, but I added more explanation in a comment as to why we're using it for re-assignment.

src/recordlinker/linking/link.py
```diff
@@ -122,54 +206,74 @@ def link_record_against_mpi(
...
+    prediction: schemas.Prediction = "possible"
     matched_person: typing.Optional[models.Person] = None
-    results: list[LinkResult] = [
-        LinkResult(k, v) for k, v in sorted(scores.items(), reverse=True, key=lambda i: i[1])
-    ]
-    result_counts["above_lower_bound"] = len(results)
-    if not results:
+
+    if not results and not certain_results:
```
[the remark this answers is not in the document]
bamader: ditto above

src/recordlinker/assets/initial_algorithms.json
```diff
@@ -25,7 +21,11 @@
                         "func": "func:recordlinker.linking.matchers.compare_probabilistic_fuzzy_match"
                     }
                 ],
-                "rule": "func:recordlinker.linking.matchers.rule_probabilistic_match",
+                "rule": "func:recordlinker.linking.matchers.rule_probabilistic_sum",
+                "possible_match_window": [
+                    0.8,
+                    0.925
```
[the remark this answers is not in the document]
bamader: My gut lol. Basically I took the true_match_threshold we used to have, found where it sat as far as points were concerned (12.2 / 13.2 ~= 0.92), then built a bubbled window around it. In the absence of testing on actual prod data, I figured basing it on something else that was originally prod-adjacent was the next best thing.

tests/unit/linking/test_link.py
bamader: Yeah that's fair. I updated the other test that uses `identifier` to follow suit, since it's ambiguous. The other tests all have numbers above 5 or 6, so I don't think normalization is as much a concern, and to write out their decimal would be calling `round( jaro-winkler-similarity( name1, name2 ) * log-odds[FIRST_NAME] ) etc` and at that point it feels redundant to me since that's just reproducing the function code.

tests/unit/linking/test_link.py
bamader: I do think this is something worth investigating. I think we'll have to base this heavily on what happens with Montana  and creating stuff for them to run. We'd have to do profiling around SSN first anyway since we didn't even have that data nicely presented before.

bamader:
Flagging that I've edited the ticket's AC due to scoping down work and the urgency of progressing this to allow our clients to test. There won't be any logical or performance based hits as a result, merely moving work to consolidate and refactor code into a follow-up.

bamader:
Follow up ticket with class consolidation: https://app.zenhub.com/workspaces/dibbs-record-linker-678acc70237f5000299a9d71/issues/gh/cdcgov/recordlinker/263

src/recordlinker/linking/link.py
```diff
@@ -183,17 +287,26 @@ def link_record_against_mpi(
                 commit=False,
             )
 
+    # Put together some strings to report the result, window, and interpretation
+    # to the user; we don't save certainly-not grades in the dynamic table
+    best_score_str: str = "n/a"
+    reference_range: str = "n/a"
+    if prediction == "certain":
+        best_score_str = str(certain_results[0].rms)
+        reference_range = "(" + str(certain_results[0].mmt) + ", " + str(certain_results[0].cmt) + ")"
```
ericbuckley, line 296:
```suggestion
        reference_range = f"({certain_results[0].mmt}, {certain_results[0].cmt})"
```

src/recordlinker/linking/link.py
```diff
@@ -30,15 +31,79 @@
...
+    to update relative match score strengths as well as prioritizing certain
+    matches over possible matches).
+    """
     person: models.Person
-    belongingness_ratio: float
+    accumulated_points: float
+    rms: float
+    mmt: float
+    cmt: float
+    grade: str
```
ericbuckley, line 47: If you make this a schema class, say `schemas.Prediction(enum.Enum)` rather than a str, we can get a couple of benefits.

1) We'll have better documentation in the API as now our users will know its always a "certainly", "possibly" or "certainly-not" enum rather than any sequence of characters.
2) We can explicitly define order on the values, that will abstract logic in handle_update.  Mainly, deciding when one  Prediction grade should override another.

bamader: Very good point about documentation, I've updated the parameter signature to correspond to `MatchGrade`'s literal typing.

As far as enums go, I left some rationale in a comment on the ticket, but I've actually decided against going the ordered enum route here. Ordering imposes an overhead in method generation, attribute access, and equality checks / comparisons, and since we perform these checks on the order of n^2 (one loop over cluster, one loop over patients in cluster) in our most commonly invoked function, that overhead seemed meaningful (confirmed in testing below). Looking at the code side by side (string comparisons vs enum abstraction), I also found making these values into an ordered enum a bit overkill when we can get by with just an equality comparison and don't leverage any other complex functionality. I like the way the update function reads as-is, because users can see the logic process of the cases where we would update a match grade, which I think is a built-in way the code can assuage some RMS concerns. 

To check on the performance implications, I ran a `timeit` test comparing enums (first result) with strict string equality (second) just to make sure things were as efficient as possible:
Screenshot 2025-03-27 at 2 20 53 PM

Unless there's a strong desire to change this anyway, I'd opt to leave it as-is.

ericbuckley (commented):
Looking good, thanks for making all the changes per the github issue so far.

src/recordlinker/schemas/link.py
```diff
@@ -41,10 +41,29 @@ class LinkResult(pydantic.BaseModel):
...
     )
-
-    belongingness_ratio: typing.Annotated[float, pydantic.Field(ge=0, le=1)] = pydantic.Field(
-        description="The proportion of patient records matched in this person cluster ("
-        "between 0 and 1.0)."
+    accumulated_points: float = pydantic.Field(
+        description="The median number of log-odds points accumulated by the blocked patients "
+        "belonging to this Person Cluster when compared to an incoming record."
+    )
+    pass_number: int = pydantic.Field(
```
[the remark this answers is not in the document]
bamader: Good call, since your label PR landed I've made the update

src/recordlinker/linking/link.py
```diff
@@ -30,8 +31,69 @@
...
+        case-based logic clearer.
+        """
+        self.accumulated_points = earned_points
+        self.pass_number = pass_num
+        self.rms = rms
+        self.grade = grade
+        self.cmt = cmt
+        self.mmt = mmt
+    
+    def check_and_update_score(
```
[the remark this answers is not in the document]
bamader: I've already addressed this in two places (comment in the ticket AC and a comment above). We care about only one of the six comparison functions, making this decorator overkill. It imposes a non-trivial performance hit as measured above to generate those methods we don't intend to use. And when comparing the look of the code side by side, I strongly prefer the more explicitly enumerated cases in an `if/elif` block as I think it provides clarity and transparent value to users reading the code--decision logic is something we have continuously seen users struggle to grasp on first read, the easier we can lay it out, the better.

m-goggins (commented):
This looks great to me! The only thing I'd add is a test to cover the lines that don't have coverage in `link.py`. Normally, I wouldn't be worried about the test coverage since it's not a big hit for those lines but I do think we want to demonstrate how the match grade is determined when the grades are not the same.

src/recordlinker/schemas/link.py
```diff
@@ -41,10 +41,29 @@ class LinkResult(pydantic.BaseModel):
...
-        "between 0 and 1.0)."
+    accumulated_points: float = pydantic.Field(
+        description="The median number of log-odds points accumulated by the blocked patients "
+        "belonging to this Person Cluster when compared to an incoming record."
+    )
+    pass_label: str = pydantic.Field(
+        description="The label associated with the algorithm pass that generated this "
+        "link result."
+    )
+    rms: typing.Annotated[float, pydantic.Field(ge=0, le=1)] = pydantic.Field(
```
ericbuckley, line 52: I just did a quick scan of the API to see what kinds of labels we have for schema properties.  Outside of `kwargs` on the algorithm pass (which is in the process of being removed), I didn't see any other places where we use acronyms for property names.

This doesn't have to be done now, but before we release v25.5.0, I think we should consider replacing these (rms, mmt, cmt) acronyms with full names.

bamader:
@m-goggins Super fair point, and I think it's important to show that grade adaptation is a new feature of the RMS decision mechanism. I've added tests to cover both directions (existing grade is certain so no change in pass 2, and existing grade is only possible so definite change in pass 2) so we should now have total coverage for our `link.py` case logic.

m-goggins (approved):
Great work, really excited to get this feature out the door to NBS! Thanks for adding the additional tests

PR #240, opened 2025-03-06
Add state normalization as part of PIIRecord construction

### LINKED ISSUE #202: parse state values
## Summary
Update the Address class to parse state values as data comes in.  Data should be normalized into standard USPS two letter alphabetic codes, this includes states, possessions and military codes.  If a state can not be accurately parsed, a validation error should be raised.

## Acceptance Criteria
- [ ] A new parse_state method on the Address class
- [ ] Test cases to show that state is correctly parsed
- [ ] Documentation added to site/references to indicate that state is normalized into the two-letter code

## Details / Tasks
A valid list of all the two-letter codes can be found here https://www.hl7.org/fhir/us/core/STU4/ValueSet-us-core-usps-state.html.

## Background / Context
State is current accepted as raw text, and saved that way for comparisons. In some cases the record might use an abbreviation and in others, the full state name.  Ideally, comparisons like "AK" <-> "Alaska", should be a 100% match.

m-goggins:
## Description
This PR adds a method on the `Address` class to normalize states in a `PIIRecord` to the 2-digit USPS code if the provided state is valid. We may want to update the other normalization functions, e.g., `phone_number` and `email` to ensure that the same normalization occurs for those fields as well.

## Related Issues
Fixes #202


src/recordlinker/schemas/pii.py
```diff
@@ -222,12 +238,7 @@ def model_construct(
         is this method will not parse and validate the data, thus should only be used
         when the data is already cleaned and validated.
         """
-        obj = super(PIIRecord, cls).model_construct(_fields_set=_fields_set, **values)
-        obj.address = [Address.model_construct(**a) for a in values.get("address", [])]
-        obj.name = [Name.model_construct(**n) for n in values.get("name", [])]
-        obj.telecom = [Telecom.model_construct(**t) for t in values.get("telecom", [])]
-        obj.identifiers = [Identifier.model_construct(**i) for i in values.get("identifiers", [])]
-
+        obj = cls(**values)
```
ericbuckley, line 241: @m-goggins I have a feeling it might be better to discuss this on a call.  The differences between `cls.model_construct()` and `cls()` come down to performance.  The former just builds the pydantic object, the latter builds and validates it.  The general idea is that we don't need to validate data in the database as it was already validated/normalized during ingestion.  I'm interested to learn more about what issues you found with this approach.

src/recordlinker/schemas/pii.py
```diff
@@ -437,3 +448,70 @@ def blocking_values(self) -> typing.Iterator[tuple[models.BlockingKey, str]]:
             # a PII data dict could have multiple given names
             for val in self.blocking_keys(key):
                 yield key, val
+
+
+_STATE_NAME_TO_CODE = {
```
ericbuckley, line 453: What do you think about moving this to `src/assets/states.json`?  We have a [function in utils/path.py](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/utils/path.py#L17) to help with loading of the data.

src/recordlinker/schemas/pii.py
```diff
@@ -163,6 +163,22 @@ class Address(pydantic.BaseModel):
...
+        """
+        Normalize the state field into 2-digit USPS code.
+        """
+        if value:
+            value = value.strip().title()
+
+            if len(value) == 2 and value.upper() in _STATE_CODE_TO_NAME:
+                return value.upper()
+
+            if value in _STATE_NAME_TO_CODE:
```
ericbuckley, line 177: I think this search also needs to be case-insensitive.  I'm thinking what if the incoming payload has `"state": "oregon"`.

src/recordlinker/schemas/pii.py
m-goggins: reverted back per our conversation

src/recordlinker/schemas/pii.py
```diff
@@ -163,6 +165,25 @@ class Address(pydantic.BaseModel):
     latitude: typing.Optional[float] = None
     longitude: typing.Optional[float] = None
 
+    @functools.cached_property
+    def normalize_state(self) -> str | None:
+        """
+        Normalize the state field into 2-digit USPS code.
+        """
+
+        _STATE_NAME_TO_CODE = utils.read_json("assets/states.json")
```
ericbuckley, line 174: Can we move these to two variables to either the module or class level.  As is, it will call this every time the function is invoked, but we really only need to call it once.

tests/unit/schemas/test_pii.py
```diff
@@ -329,6 +329,24 @@ def test_feature_iter_given_name(self):
...
         ) == ["John", "Jon"]
 
+    def test_feature_iter_state(self):
+        record = pii.PIIRecord(
+            address=[
+                pii.Address(state="new york"),
+                pii.Address(state="NY"),
+                pii.Address(state="California"),
+                pii.Address(state=" california "),
+                pii.Address(state="of mind"),
```
ericbuckley, line 339: 🤣

src/recordlinker/schemas/pii.py
```diff
@@ -163,6 +169,23 @@ class Address(pydantic.BaseModel):
     latitude: typing.Optional[float] = None
     longitude: typing.Optional[float] = None
 
+    @functools.cached_property
+    def normalize_state(self) -> str | None:
+        """
+        Normalize the state field into 2-digit USPS code.
```
ericbuckley, line 175:
```suggestion
        Normalize the state field into 2-letter USPS code.
```

ericbuckley (approved):

Issue #231, opened 2025-02-28
improve blocking by exluding incorrect matches


ericbuckley:
## Summary
The blocking algorithm is too liberal in including **all patients** from overlapping person clusters.  We should include those patients that don't match because of missing key values, but we should exclude patients with different values.

## Acceptance Criteria
- [ ] re-implement the `get_block_data` method to exclude patients that don't match the blocking criteria
- [ ] new test cases to show that patients with incorrect matches are being excluded

## Details / Tasks
There are likely two ways to solve this, a) modify the query b) keep the query as is and filter the dataset in memory.  It's not obvious which strategy is better from a performance standpoint, please consider both when implementing and include some documentation / comments on why that strategy was taken.

## Background / Context
Take a simple blocking example of finding all patients with a matching BIRTHDATE of '1999-01-01'.  Currently, we find all patient records in the database with that matching BIRTHDATE in the blocking values table, we then reduce that to a unique set of person clusters (based on what clusters those records belong to).  We then return **all patients** in those matching person clusters, this creates 3 groups of data in the result set.
- Patients that exactly match on '1999-01-01'
- Patients that have no BIRTHDATE value
- Patients that have a BIRTHDATE value, but does not match '1999-01-01'

The last group in the result set is problematic, it clearly doesn't block, and therefore shouldn't be present in our evaluation.  Removing this group from our result set, should provide more accurate calculations on cluster membership.  

## Related Issues/PRs
#230

ericbuckley:
@bamader @m-goggins two questions for you both

1) Does the description above match your expectations from our discussion this week?
2) Does this change benefit BOTH the belongingness ratio algorithm and the weighted confidence one we're working on?  If it's only the latter, I only want to implement this work post our weighted confidence implementation.

PR #242, opened 2025-03-10
Handle missing values in blocked MPI records

bamader:
## Description
This PR adds a filter function to the `get_block_data` function in the `mpi_service`. This filter function removes from consideration all candidates from the MPI who are part of a person cluster containing a patient who satisfied blocking criteria, but which themselves have present values (i.e. are not missing fields) that disagree with incoming blocking keys. This will make matching more precise and reinforce validity in the blocking step.

## Related Issues
#231 

## Additional Notes
N/A

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/database/mpi_service.py
+++ b/src/recordlinker/database/mpi_service.py
@@ -18,6 +18,64 @@ from recordlinker import models
 from recordlinker import schemas
 
 
+def _filter_incorrect_blocks(
+        record: schemas.PIIRecord,
+        patients: typing.Sequence[models.Patient],
+        blocking_keys: list[str]
+) -> list[models.Patient]:
+    """
+    Filter a set of candidates returned via blocking from the MPI. The initial
+    SQL query returns a collection of candidates comprised of all patients in
+    the MPI belonging to a Person cluster for which at *least one* patient
+    satisfied blocking criteria. This function filters that candidate set to
+    include *only* those patients who either satisfied blocking criteria or
+    were missing a value for one or more blocked fields. This eliminates 
+    patients from consideration who have mismatched blocking information but
+    belonged to a Person cluster where a different record had correct blocking
+    values.
+
+    :param record: The PIIRecord of the incoming patient.
+    :param patients: The initial collection of candidates retrieved from the MPI.
+    :param blocking_keys: A list of strings of the fields used for blocking.
+    :returns: A filtered list of Patients from the MPI.
+    """
+    # Extract the acceptable blocking values from the incoming record
+    # Keys have already been getattr validated by caller, no need
+    # to check that they exist
+    blocking_vals_in_incoming = {}
+    for bk in blocking_keys:
+        key = getattr(models.BlockingKey, bk)
+        blocking_vals_in_incoming[bk] = [v for v in record.blocking_keys(key)]
+
+    # Can't modify sequence in place, so we'll build up a list of list idxs
+    # to exclude for mpi patients who don't match blocking criteria exactly
+    pats_to_exclude = set()
+    print("incoming blocking vals", blocking_vals_in_incoming)
+    for p in patients:
+        num_agreeing_blocking_fields = 0
+        mpi_record = p.record
+        for bk, allowed_vals in blocking_vals_in_incoming.items():
+            # Compare incoming blocking value to what would be the blocking
+            # value of the mpi record to make sure we compare on e.g. same
+            # number of characters at beginning/end of string
+            mpi_vals = mpi_record.blocking_keys(getattr(models.BlockingKey, bk))
+
+            # Generator gets us best performance, fastest way to check membership
+            # because we return True as soon as we get 1 rather than build the
+            # whole list. Also count compatibility if mpi_val is missing.
+            found_compatible_val = (len(mpi_vals) == 0) or any(x in mpi_vals for x in allowed_vals)
+            if found_compatible_val:
+                num_agreeing_blocking_fields += 1
+
+        # If we get through all the blocking criteria with no missing entries
+        # and no true-value agreement, we exclude
+        if num_agreeing_blocking_fields < len(blocking_keys):
+            print("Found patient to exclude", p.data)
+            pats_to_exclude.add(p.id)
+    
+    return [pat for pat in patients if pat.id not in pats_to_exclude]
+
+
 def get_block_data(
     session: orm.Session, record: schemas.PIIRecord, algorithm_pass: models.AlgorithmPass
 ) -> typing.Sequence[models.Patient]:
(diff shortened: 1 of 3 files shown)


src/recordlinker/database/mpi_service.py
```diff
@@ -18,6 +18,64 @@
...
+        for bk, allowed_vals in blocking_vals_in_incoming.items():
+            # Compare incoming blocking value to what would be the blocking
+            # value of the mpi record to make sure we compare on e.g. same
+            # number of characters at beginning/end of string
+            mpi_vals = mpi_record.blocking_keys(getattr(models.BlockingKey, bk))
+
+            # Generator gets us best performance, fastest way to check membership
+            # because we return True as soon as we get 1 rather than build the
+            # whole list. Also count compatibility if mpi_val is missing.
+            found_compatible_val = (len(mpi_vals) == 0) or any(x in mpi_vals for x in allowed_vals)
```
ericbuckley, line 66: If you do the reverse here, which is to look for an incompatible value.  You can short-circuit the for loop and break out as soon as you find the first incompatible value.  No need to check every key.

src/recordlinker/database/mpi_service.py
```diff
@@ -18,6 +18,64 @@
...
+    :param blocking_keys: A list of strings of the fields used for blocking.
+    :returns: A filtered list of Patients from the MPI.
+    """
+    # Extract the acceptable blocking values from the incoming record
+    # Keys have already been getattr validated by caller, no need
+    # to check that they exist
+    blocking_vals_in_incoming = {}
+    for bk in blocking_keys:
+        key = getattr(models.BlockingKey, bk)
+        blocking_vals_in_incoming[bk] = [v for v in record.blocking_keys(key)]
```
ericbuckley, line 48: I think in the case where the valueset is empty, we want to exclude it from the dict.

src/recordlinker/database/mpi_service.py
ericbuckley: First, thanks for profiling, and if you can specifically show with real or synthetic data that one test is faster than the other, clearly let's go that way.

Also, I recognize what you're saying about the two loops (the blocking keys and the allowed values), but I believe some of what you're saying assumes the cost to generate the lists, blocking keys and allowed values, is equal.  The blocking keys are precomputed on lines 46-48, and in some ways are required for us to do this (really depending on where this logic lives, it's not, because this logic has already been performed on lines 102, a different solution say using an inner function could reuse).  However, the inner loop is subject to compute time necessary for calculating the mpi_values on line 61.  This is fast, but there is some cost to running the feature iterators and normalization functions.

ericbuckley: To elaborate more on what I mean about inner function that could reuse, take for example this patch

```patch
diff --git i/src/recordlinker/database/mpi_service.py w/src/recordlinker/database/mpi_service.py
index 56f1fcd..631f676 100644
--- i/src/recordlinker/database/mpi_service.py
+++ w/src/recordlinker/database/mpi_service.py
@@ -85,9 +85,20 @@ def get_block_data(
     remaining Patient records in the Person clusters identified in
     blocking to calculate Belongingness Ratio.
     """
+    def _filter_incorrect_blocks(
+        blocking_values: dict[models.BlockingKey, set[str]], record: schemas.PIIRecord
+    ) -> bool:
+        for key, vals in blocking_values.items():
+            if vals:
+                record_vals = set(record.blocking_keys(key))
+                if record_vals and (not record_vals & vals):
+                    return False
+        return True
     # Create the base query
     base = expression.select(models.Patient.person_id).distinct()
 
+    blocking_values: dict[models.BlockingKey, set[str]] = {}
+
     # Build the join criteria, we are joining the Blocking Value table
     # multiple times, once for each Blocking Key.  If a Patient record
     # has a matching Blocking Value for all the Blocking Keys, then it
@@ -99,7 +110,7 @@ def get_block_data(
         key = getattr(models.BlockingKey, key_id)
 
         # Get all the possible values from the data for this key
-        vals = [v for v in record.blocking_keys(key)]
+        blocking_values[key] = {v for v in record.blocking_keys(key)}
         # Create a dynamic alias for the Blocking Value table using the index
         # this is necessary since we are potentially joining the same table
         # multiple times with different conditions
@@ -112,14 +123,14 @@ def get_block_data(
             expression.and_(
                 models.Patient.id == alias.patient_id,
                 alias.blockingkey == key.id,
-                alias.value.in_(vals),
+                alias.value.in_(blocking_values[key]),
             ),
         )
 
     expr = expression.select(models.Patient).where(models.Patient.person_id.in_(base))
     candidates = session.execute(expr).scalars().all()
-    return _filter_incorrect_blocks(record, candidates, algorithm_pass.blocking_keys)
+    return filter(functools.partial(_filter_incorrect_blocks, blocking_values), candidates)
 
 
 def insert_patient(
```

bamader: I think that's a super fair point about cost amortization and compute time expenditures on cleansing functions, especially with the normalization work Marcelle will be doing. It looks to me like there are reasonable performance considerations on both sides of the equation here, and it's probably the case we won't know for sure which one will scale truly optimally on production data because we just don't have an idea of the lengthening of the data norms. Since the `timeit` values are so close and we have sort of counter-balancing factors for searching both compatible and incompatible values, I think in the end I ultimately don't have a preference for which search method we take. I do think I have a pretty firm preference to stick with a separate function and incorporate the logging information you mention in another comment, rather than internally sub-defining as in the patch, so in light of probably what's going to be comparable / a wash performance-wise, would you prefer I implement the compatible values loop or incompatible values loop?

ericbuckley: Nope, I just wanted you to consider it.  Please go with what you think reads better if you've shown the performance differences between the two are negligible.

bamader: Sounds good. I like the separate function I've written to search for compatible values, so I'll stick with that for now, but I'll also leave a comment explaining the justification and flagging that in the future if preprocessing functions end up a bigger time cost it could be worth switching the preference of which loop to short circuit.

src/recordlinker/database/mpi_service.py
bamader: Yeah I think that's right. It starts to touch a little bit on the other half of missing information, which is missing fields in the incoming record (to be handled by our log odds wizardry), but since we'll be implementing that elsewhere this function won't need to be aware of missing incoming information; it can safely assume that if it's being run, the caller was satisfied that there's enough content to execute blocking.

src/recordlinker/database/mpi_service.py
```diff
@@ -18,6 +18,64 @@
...
+
+    # Can't modify sequence in place, so we'll build up a list of list idxs
+    # to exclude for mpi patients who don't match blocking criteria exactly
+    pats_to_exclude = set()
+    print("incoming blocking vals", blocking_vals_in_incoming)
```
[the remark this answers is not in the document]
bamader: Turns out we can't log this information because what's most helpful for logging are the actual field values used in blocking, but that's PII so we can't leak that. It's safer to just scrap the print/logs after all.

src/recordlinker/database/mpi_service.py
bamader: Per the conversation, noting the approach discussion in a comment and indicating we're sticking with compatible search.

ericbuckley (approved):

Issue #237, opened 2025-03-06
ambiguous birthdate years should not be future dates


ericbuckley:
## Summary
When parsing birthdates with ambiguous years, eg 2/19/40, we should assume it's a date in the past (1940) and not the future (2040).

## Impact
Incorrectly parsing the date leads to poor BIRTHDATE matches when comparing records.

## Expected behavior
When a date is parsed with a 2 digit year, we should always assume it's in the past.  `2/19/40` == `Feb 19th 1940`.  When possible, we should assume the current century and not every 2-digit year should be interpreted as the 20th century.  For example, `5/15/04` == `May 15th 2004` and **not** `May 15th 1904`.  Additionally, we should be rejecting all values that are clearly future dates, eg `2050-01-01` or dates that are more than 150 years old eg `1885-01-01`.

bamader:
I recommend an important clarification here: we definitely want two-digit years to be in the past, but I did some research and it seems to be normatively common to choose a century based off the current year. If we get a date after `25`, we logically assume it's `19XX`. But if we receive something like `08`, it seems to be accepted practice to interpret that as `2008` rather than `1908`. This also makes more logical sense to me: we can realistically expect to receive health records from someone born 17 years ago, not so much 117 years.

ericbuckley:
> I recommend an important clarification here: we definitely want two-digit years to be in the past, but I did some research and it seems to be normatively common to choose a century based off the current year. If we get a date after `25`, we logically assume it's `19XX`. But if we receive something like `08`, it seems to be accepted practice to interpret that as `2008` rather than `1908`. This also makes more logical sense to me: we can realistically expect to receive health records from someone born 17 years ago, not so much 117 years.

I agree, I'll update the description with more info.

PR #245, opened 2025-03-10
Parse and validate two-digit dates

bamader:
## Description
This PR does two things:
1. Adds more robust-handling of two-digit date strings, so that dates like `02/09/39` parse as 1939 rather than 2039,
2. Adds validation that sets boundaries around acceptable dates by not accepting dates before 1850 and after the current day.

## Related Issues
#237 

## Additional Notes
N/A

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -4,8 +4,7 @@ import json
 import re
 import typing
 
-import dateutil.parser
-from dateutil.parser import parserinfo
+from dateutil.parser import parse, parserinfo
 import pydantic
 
 from recordlinker import models
@@ -262,7 +261,7 @@ class PIIRecord(pydantic.BaseModel):
                         year -= 100
                 return year
         if value:
-            given_date = dateutil.parser.parse(str(value), LinkerParserInfo())
+            given_date = parse(str(value), LinkerParserInfo())
             if given_date > datetime.datetime.today():
                 raise ValueError("Dates cannot be in the future")
             if given_date < datetime.datetime(1850, 1, 1):
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -247,11 +247,18 @@ class PIIRecord(pydantic.BaseModel):
         class LinkerParserInfo(parserinfo):
             def convertyear(self, year, *args):
                 """
-                Subclass method override for parser info function dedicatedo to
-                handling two-digit year strings.
+                Subclass method override for parser info function dedicated to
+                handling two-digit year strings. The Parser interprets any two
+                digit year string up to and including the last two digits of
+                the current year as the current century; any two-digit value 
+                above this number is interpreted as the preceding century.
+                E.g. '25' is parsed to '2025', but '74' becomes '1974'.
+                The Parser does not accept dates in the future, even if in 
+                the same calendar year.
                 """
                 # self._year is the current four-digit year
-                # self._century is leading two digits of self._year
+                # self._century is self._year with the tens and ones digits dropped, e.g.
+                # 19XX becomes 1900, 20XX becomes 2000
                 # implementation override follows template pattern in docs
                 # https://dateutil.readthedocs.io/en/latest/_modules/dateutil/parser/_parser.html#parserinfo.convertyear # noqa: E712
                 if year < 100:
@@ -264,9 +271,9 @@ class PIIRecord(pydantic.BaseModel):
         if value:
             given_date = parse(str(value), LinkerParserInfo())
             if given_date > datetime.datetime.today():
-                raise ValueError("Dates cannot be in the future")
+                raise ValueError("Birthdates cannot be in the future")
             if given_date < datetime.datetime(1850, 1, 1):
-                raise ValueError("Dates cannot be before 1850")
+                raise ValueError("Birthdates cannot be before 1850")
             return given_date
 
     @pydantic.field_validator("sex", mode="before")
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -5,6 +5,7 @@ import re
 import typing
 
 import dateutil.parser
+from dateutil.parser import parserinfo
 import pydantic
 
 from recordlinker import models
@@ -243,8 +244,30 @@ class PIIRecord(pydantic.BaseModel):
         """
         Parse the birthdate string into a datetime object.
         """
+        class LinkerParserInfo(parserinfo):
+            def convertyear(self, year, *args):
+                """
+                Subclass method override for parser info function dedicatedo to
+                handling two-digit year strings.
+                """
+                # self._year is the current four-digit year
+                # self._century is leading two digits of self._year
+                # implementation override follows template pattern in docs
+                # https://dateutil.readthedocs.io/en/latest/_modules/dateutil/parser/_parser.html#parserinfo.convertyear # noqa: E712
+                if year < 100:
+                    year += self._century
+                    if year > self._year:
+                        # This allows us to continually make a pivot at the current year;
+                        # Keeps with best practice and conventional norms
+                        year -= 100
+                return year
         if value:
-            return dateutil.parser.parse(str(value))
+            given_date = dateutil.parser.parse(str(value), LinkerParserInfo())
+            if given_date > datetime.datetime.today():
+                raise ValueError("Dates cannot be in the future")
+            if given_date < datetime.datetime(1850, 1, 1):
+                raise ValueError("Dates cannot be before 1850")
+            return given_date
 
     @pydantic.field_validator("sex", mode="before")
     def parse_sex(cls, value):
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -3,8 +3,8 @@ import enum
 import json
 import re
 import typing
-import pydantic
 
+import pydantic
 from dateutil.parser import parse
 from dateutil.parser import parserinfo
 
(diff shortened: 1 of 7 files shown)


tests/unit/assets/patient_bundle_to_link_with_mpi.json
```diff
@@ -31,7 +31,7 @@
             "use": "official"
           }
         ],
-        "birthDate": "2053-11-07",
+        "birthDate": "2013-11-07",
```
ericbuckley, line 34: wow, I didn't realize our tests had so many invalid birth dates!

tests/unit/schemas/test_pii.py
```diff
@@ -101,10 +101,20 @@ def test_parse_birthdate(self):
         assert record.birth_date == datetime.date(1980, 1, 1)
         record = pii.PIIRecord()
         assert record.birth_date is None
+        record = pii.PIIRecord(birthdate="02-09-39")
+        assert record.birth_date == datetime.date(1939, 2, 9)
```
ericbuckley, line 105: nitpick; I'm wondering if we should just delete this assertion.  I believe it's the testing the same assertion on line 107, and it will fail in 14 years (the one on 107 fails in 49 years).  14 years is a long time from now, but I want to think this code will still be running then 😂 .  

Honestly, not a big deal, just thinking it doesn't add much additional value over 107.

src/recordlinker/schemas/pii.py
```diff
@@ -243,8 +244,30 @@ def parse_birth_date(cls, value):
...
+                    if year > self._year:
+                        # This allows us to continually make a pivot at the current year;
+                        # Keeps with best practice and conventional norms
+                        year -= 100
+                return year
         if value:
-            return dateutil.parser.parse(str(value))
+            given_date = parse(str(value), LinkerParserInfo())
+            if given_date > datetime.datetime.today():
+                raise ValueError("Dates cannot be in the future")
```
ericbuckley, line 267:
```suggestion
                raise ValueError("Birthdates cannot be in the future")
```

src/recordlinker/schemas/pii.py
```diff
@@ -243,8 +244,30 @@ def parse_birth_date(cls, value):
...
+            if given_date < datetime.datetime(1850, 1, 1):
+                raise ValueError("Dates cannot be before 1850")
```
ericbuckley, line 269:
```suggestion
                raise ValueError("Birthdates cannot be before 1850")
```

src/recordlinker/schemas/pii.py
```diff
@@ -243,8 +244,30 @@ def parse_birth_date(cls, value):
...
+                Subclass method override for parser info function dedicatedo to
+                handling two-digit year strings.
+                """
+                # self._year is the current four-digit year
+                # self._century is leading two digits of self._year
+                # implementation override follows template pattern in docs
+                # https://dateutil.readthedocs.io/en/latest/_modules/dateutil/parser/_parser.html#parserinfo.convertyear # noqa: E712
+                if year < 100:
+                    year += self._century
+                    if year > self._year:
```
ericbuckley, line 259: I'm thinking this block will fail if the date passed in is, `2025-11-20`.  A future date, this calendar year.

tests/unit/schemas/test_pii.py
bamader: Yeah that's totally fair. The only reason this is in here is because it's the date of a specific edge case in the NBS exception set, but the concept is already sufficiently tested. I don't think we need this one.

src/recordlinker/schemas/pii.py
ericbuckley: Oh I think I got this backwards, let me say it in another way.  If the current date is `2025-11-20` and we receive a birthdate for `1/10/25`, will this block convert that to Jan 10th 1925 or Jan 10th 2025?

bamader: Ohhh I see what you're saying. I think in this case we'd be okay though. If we received a date of `1/10/25`, the year would be `25` and thus less than 100, so we'd add century (`2000`) to it, making it 2025. 2025 is not greater than the internal `2025` that is `self._year`, so we'd return the 2025 parsing. However, what your question _does_ highlight is that my documentation clarification for `self._century` is incorrect--it's not just the leading two digits, it's actually the year rounded down a century, treating the tens and ones places as a "remainder". So any date that starts with 19XX will have `self._century = 1900` and any date like `20XX` will have `self._century = 2000`. This is worth clarifying and might address the confusion.

ericbuckley: Let me just write it as a test case, when I add the following to the `test_parse_birthdate` function it fails.

```python
        record = pii.PIIRecord(birthdate="12-19-25")
        assert record.birth_date == datetime.date(1925, 12, 19)
```

bamader: I'm not understanding what you're trying to say. I think this can be simplified to two decisions about desired behavior.
1. When the system receives an ambiguous century date marked `25`, do we want that parsed as 1925 or as 2025? I very strongly believe it should be 2025 because that is an extremely common convention among US date reporting. This is why the test case you wrote fails; `25` gets parsed as `2025` which is imo correct.
2. When the system receives a date that is in the future relative to today, but is still in the same calendar year as today, should the system accept that date? I would propose no because clinical health data should not have recorded observations (which in medical terminology denote events that have been fully processed and are being sent for reporting) for things that have not happened yet.

ericbuckley: 1.  I think the both of us are interpreting the requirements differently.  When we spoke about this the other week, my interpretation of the problem was an **ambiguous future birthdate** should be assumed to have taken place 100 years ago.  It sounds like your interpretation of the problem is an **ambiguous future birth year** should be assumed to have taken place 100 years ago.   So the difference being are we comparing birth year or birthdate.  I'm very interested though in what you mean by extremely common convention among US date reporting.  I haven't done any reading on this, could you share more about where you've seen similar corrections made?
2. Yes.  However, a) we only accept 1 date time value, the birthdate, all other date time values received are ignored.  b) we're going to [remove data cleansing as a whole](https://github.com/CDCgov/RecordLinker/issues/248) (well everything except removing leading/trailing whitespace) in the near future.  The idea being, that we should preserve the raw data and just normalize before comparisons.  Meaning after #248 is completed, we'll accept any date value, but won't emit a birth date for comparisons if its in the future.

bamader: 1. I think the conversation is moot. The `convertyear` function in the `ParserObject`, which is what we need to override to implement manual century checks, doesn't receive any info about the date other than the year. See the source [here](https://dateutil.readthedocs.io/en/stable/_modules/dateutil/parser/_parser.html#parserinfo.convertyear). That means we can't make comparisons to the month and day when arbitrating a year, and need to restrict to only looking at century. In this case, I think we should just go with what I have. More generally, I was referring to it being a common thing to just express a date year as two-digit. There's no study or source on that, it's just something people commonly do. I've seen it on checks, bank documents, employment paperwork, medical charts. Not really sure how to "prove" that's a social convention.
2. For purposes of this work I can remove throwing the error on a future date, but did we at some point talk about removing preprocessing prior to ingestion? That makes no sense to me as something desirable to do. Re-applying pre-processing every time linkage is run is an unnecessary performance hit. I'm also not sure I believe that maintaining the absolutely original form of received data has a meaningful benefit, as opposed to standardized and processed data. DIBBs did the latter to great success in prior projects, I'm not sure why we'd change that. But, in either case, it doesn't really affect this work, so I can drop the future date error.

bamader: Coming back to this--where did we land after the midday call? Number 1 kind of is what it is because of the function signature, so that'll be staying as is, but for 2, before I dropped I think we decided we did want to reject future dates after all? That might have changed though so wanted to follow up.

ericbuckley: 1. FWIW, we have options outside of `ParserObject`, for one we don't need to use dateutil.  But that's implementation specific, and not really what we're discussing anyway.  My point from above was that we were on different pages.  You assumed, and you're not wrong, that the Github issue asked to convert **future birth years** to 19XX dates, I assumed it was **future birth dates**.  I'm fine leaving it the way you have it, but I'd appreciate it if you added some extra documentation to site/reference.md explaining how ambiguous dates are handled.  Doesn't need to be a lot, but just enough to say we make our best attempt to process ambiguous birthdates but will fail if any date resolves to a future one.
2. It's fine to reject future birthdates here.  #248 needs some work, so we can handle any cleansing vs normalization changes when that issue is ready to go.

bamader: I did assume, you're right. To me, it seems a more likely scenario that someone would write '25' to refer to the current year they're living in rather than '1925', but re-reading the ticket I see that isn't the precise ask. I tend to view tickets as placeholders for a type of work rather than as immutable or something a PR is bound to more than loosely, if an engineer thinks there's room to further improve or needs to exercise judgment. But this isn't a hill I wish to die on, so I apologize, and I'll stick to exact wording going forward. 

For this specific case, I also assumed we were in fact bound to `dateutil`. When we developed record linkage for LAC, we had different date validation, so I assumed that since it was replaced with `dateutil` and `ParserObject` there was a compelling reason that change had been made and so we wanted to stick within that ecosystem (hence why I argued from the position of the existing function signature). If this isn't the case, I'm fine using something that can parse any future date as 100 years ago.

ericbuckley: No worries.  I think this implementation is fine, but would just like to see a small update to the [site/reference.md](https://github.com/CDCgov/RecordLinker/blob/main/docs/site/reference.md?plain=1#L16) to note that we do our best attempt at resolving ambiguous dates, but will fail when a birthdate received is in the future.

ericbuckley (approved):
Thanks @bamader

Issue #243, opened 2025-03-10
Similarity comparisons should ignore whitespace characters


m-goggins:
### Summary
We want to compare only the meaningful, rich parts of strings, i.e., non-whitespace characters, so they should be removed from string in `feature_iter`. 

### Impact
An example demonstrating how removing whitespace from the middle of strings produces more accurate string similarity metrics because the spaces no longer count as similar characters.

<img width="1150" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd3N3Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1664b8324ab0e6d4d7e38494a0f8ecebb7a27141/Screenshot%202025-03-10%20at%202.37.27%E2%80%AFPM.png" alt="Screenshot 2025-03-10 at 2.37.27 PM.png" />

This ticket is closely related to #238

Issue #238, opened 2025-03-06
Similarity comparisons should ignore case and non alphanumeric characters


ericbuckley:
## Summary
When comparing a feature on two different records, differences in cases and punctuation shouldn't negatively impact the score.

## Impact
Not normalizing the strings before JaroWinkler/Levenshtein comparisons negatively drops scores, when the reality is they are referencing the same value.

## Steps to reproduce
Examples to consider:
- `Thomas` vs `thomas`
- `O'Hara` vs `Ohara` 
- `321 Main St` vs `321 Main St.`
- `Jose` vs `José`
- ` Albany` vs `Albany`

## Expected behavior
The above 5 cases should result in similarity scores of 1.0.

Issue #239, opened 2025-03-06
Normalize address line suffixes


ericbuckley:
## Summary
Address suffixes commonly have many different forms, normalizing these suffixes during our validation phase will improve scores.

## Acceptance Criteria
- [ ] A new Address.line validation function to normalize suffixes
- [ ] Update site/reference.md to indicate ADDRESS suffixes are being normalized

## Details / Tasks
When parsing address lines, normalize all suffixes using the [USPS street suffix map](https://pe.usps.com/text/pub28/28apc_002.htm).  For example, `12 Main St`, `12 Main St.` and `12 Main Street` should all be normalized to `12 Main ST`.

PR #247, opened 2025-03-11
Normalize strings for comparison

m-goggins:
## Description
This PR improves the way that RecordLinker normalizes strings for comparison in `feature_iter`. It adds the `normalize_text` utility function to remove non-alphanumeric characters, convert to lowercase, and remove all whitespace (trailing, leading, and internal), adds field_validator to ensure all string fields remove trailing and leading whitespace in the `PIIRecord`, and updates `feature_iter` to more robustly apply normalization to fields.

## Related Issues
#238 
#243


src/recordlinker/schemas/pii.py
```diff
@@ -357,31 +372,40 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
             for address in self.address:
                 if address.postal_code:
                     # only use the first 5 digits for comparison
-                    yield address.postal_code[:5]
+                    yield normalize_text(address.postal_code)[:5]
```
ericbuckley, line 375: If we were to make removing of leading/trailing whitespace a normal part of the PIIRecord field validation process, we wouldn't need to call normalize_text here right?

m-goggins: You're right, we wouldn't need to use `normalize_text` to remove the whitespace here.

ericbuckley: To elaborate more on that, I was wondering if it be good to implement something like the following.  Then have all our pii.py classes inherit from it.

```python
class StrippedBaseModel(BaseModel):
    @field_validator('*', mode='before')
    def strip_whitespace(cls, v):
        if isinstance(v, str):
            return v.strip()
        return v
```

src/recordlinker/utils/normalize.py
```diff
@@ -0,0 +1,14 @@
...
+
+def normalize_text(text: str) -> str:
+    """
+    Normalize text for comparison by removing non-alphanumeric characters, converting
+    to lowercase, and removing all whitespace (trailing, leading, and internal).
+    """
+    text = "".join(
+        c for c in unicodedata.normalize("NFKD", text) if unicodedata.category(c) != "Mn"
+    )
+    text = re.sub(r"[^a-zA-Z0-9]", "", text)
```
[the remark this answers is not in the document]
ericbuckley: yeah I agree with this, I think at the very least you'd want to compile the pattern at the module level and reuse the pattern rather than compile everytime the function is called.  Another alternative to what Brandon suggested is to use [str.isalnum](https://docs.python.org/3/library/stdtypes.html#str.isalnum).

```python
return "".join(c.lower() for c in text if c.isalnum())
```

src/recordlinker/utils/normalize.py
```diff
@@ -0,0 +1,14 @@
...
+import unicodedata
+
+
+def normalize_text(text: str) -> str:
+    """
+    Normalize text for comparison by removing non-alphanumeric characters, converting
+    to lowercase, and removing all whitespace (trailing, leading, and internal).
+    """
+    text = "".join(
+        c for c in unicodedata.normalize("NFKD", text) if unicodedata.category(c) != "Mn"
```
ericbuckley, line 11: nitpick; I honestly don't know which is better, but I think converting to ascii and back would work as well.  Just adding this as something to think about, on the surface I don't see anything wrong with this implementation.

```python
text = unicodedata.normalize("NFKD", text).encode('ASCII', 'ignore').decode('ASCII')
```

src/recordlinker/utils/normalize.py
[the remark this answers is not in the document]
m-goggins: Looking at the test cases in `test_normalize`, `alnum` and `makestrans` perform identically and better than regex (as expected) and in an additional case with a much longer string, regex outperforms `alnum` and `maketrans` (the latter two still performing the same).

image

I find alnum slightly more intuitive so I'm going to implement that.

src/recordlinker/utils/normalize.py
m-goggins: Interestingly, the ASCII approach appears to be generally a slower, particularly for the address example that does not contain any accents.

image

src/recordlinker/schemas/pii.py
```diff
@@ -49,7 +49,18 @@ def __str__(self):
         return self.value
 
 
-class Feature(pydantic.BaseModel):
+class StrippedBaseModel(pydantic.BaseModel):
```
ericbuckley, line 52: Awesome, thanks for doing this ❤️

src/recordlinker/utils/normalize.py
ericbuckley: In either case, it's negligible, so really I'm not too concerned.  But I think the ASCII version is a power of 10 smaller than the original for each test cases (Note the differences in `e-05` vs `e-06`)

src/recordlinker/utils/normalize.py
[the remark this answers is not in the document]
ericbuckley: Thanks for all the performance testing work!

src/recordlinker/utils/normalize.py
m-goggins: oof that's what multi-tasking will get me 😓

src/recordlinker/schemas/pii.py
m-goggins: Yep, in agreement!

tests/unit/utils/test_normalize.py
```diff
@@ -0,0 +1,13 @@
+from recordlinker.utils import normalize as utils
+
+
+class TestNormalize:
+    def test_normalize(self):
+        text = " José O'Hara"
+        assert utils.normalize_text(text) == "joseohara"
+
+        text = "321 Main St."
+        assert utils.normalize_text(text) == "321mainst"
```
ericbuckley, line 10: I really like normalizing the non-ascii characters and the removal of the non-alphanumerics, but I'm having second thoughts on the whitespace.  I believe removing internal whitespace is going to expose us to some other edge cases, and doing so is setting us up for a tradeoff, so I want to make sure that we're balancing this correctly.

For example, we are planning to do some work that will [normalize street suffixes](https://github.com/CDCgov/RecordLinker/issues/239).  Take these 2 address `123 Heart Highway` and `123 Hearth Way`.  After our st suffix normalization work, this will result in `123 Heart HWY` and `123 Hearth WY`, but then after this normalization process we are now at `123hearthwy` and `123hearthwy`.

There is no doubt in my mind that the above is an edge case, but I believe removing internal whitespace is also an edge case and that most of the time the whitespace in addresses or names will lead to accurate comparisons.  Am I missing something?

cc @bamader @m-goggins

m-goggins: My initial thinking is that Jaro-Winkler will likely produce similar enough results in the case of street suffixes that it won't matter if we remove the internal whitespace or not because our address threshold is set to 0.9. Additionally, Jaro-Winkler gives more weight to matches that agree on the first couple of characters so the suffixes won't play as big of a role.

Here's a worst case scenario example where the address could potentially be short enough that J-W is considering the suffix and the similarity score is still greater than the threshold of 0.9. The addresses are `1 u HWY` and `1 uh WY`. 


image


The one area where it would matter a great deal is if a user chose to use the first 4 characters of an address in blocking. I only bring this example up because LAC used this approach in their algorithm. Here the addresses are very different and produce very different results if we include whitespace from the results if we remove whitespace.


image


I can't say for certain which one is more of an edge case but given that removing white space doesn't seem to impact scenario 1 (suffixes normalized) but does change the outcome for something we've seen in the wild, I'm inclined to continue to remove the whitespace.

bamader: Yeah I think Marcelle has it spot on here. That worst case scenario is a perfect illustration of why imo we should be removing white spaces. Even in the strongest instance (single character) it doesn't make enough of a difference. I also ran the Heart vs Hearth case since it's more representative on average, and in the normalized non-whitespace case compared to the normalized without removing whitespace case, both jaro winklers start with 0.98. It's only in the thousandths place that they begin to differ.

Also, I would actually argue that removing whitespace should _improve_ blocking. If we do this from the start, we're liable to filter out wrongly matching patients before they ever reach cluster membership because of an address they don't have. In the case Marcelle demonstrated, people who live at those two very different addresses shouldn't link together. Using the whitespace allows them to block together, which reduces performance at least a little bit in time and runs the risk of a faulty comparison. But throwing the space out because it isn't adding information will lead to an incorrect record being excluded where it shouldn't be cropping up.

I think removing whitespace is the right call here and will align better with real world data.

ericbuckley: OK, thank you both.  I just wanted to make sure we thought through the edge cases it didn't necessarily improve.

ericbuckley (approved):
🚀

PR #249, opened 2025-03-13
remove kwargs

ericbuckley:
## Description
Removing kwargs from Algorithm, in favor of explicit definitions of the necessary fields.  Additionally, a large refactor of the Algorithm parameter sent to linkage, this is now an instance schema.Algorithm not model.Algorithm.

## Related Issues
closes #223 

## Additional Notes
- **Evaluation Context** a new section in the Algorithm configuration was created for capturing context parameters.  This is very similar to how the existing kwargs functioned, but a) adds structure b) is defined once per Algorithm rather than once per pass.
- **Removing rule function** since the removal of dibbs-basic, there is only one evaluation rule.  It's been removed from the configuration, since it can no longer be configured.
- **Making evaluator function name easier** the inclusion of the "func:..." syntax was to allow for some flexibility in testing the phdi code against RecordLinker.  That flexibility is no longer needed, and using an enum name rather than a function path string seems easier for users to grasp. (ie, users should have to know or care where the function lives in the codebase).
- **Swap models.Algorithm for schemas.Algorithm** Theres a one way relationship between schemas and models, the former knows about the latter, but not the other way around (necessary to prevent circular imports).  This has created some problems when passing models.Algorithm around in the linkage code, for instance we always have to convert a blocking key id/string to a BlockingKey object for evaluation, same goes for Feature.   The schemas class has everything we need for parsing the data into the appropriate objects for evaluation, it made a lot of sense and reduced some of the logic needed in `recordlinker/linking` by making this switch.
- **Removing models.AlgorithmPass**. Some of the validation logic was being duplicated in the schemas.Algorithm and models.Algorithm, additionally, any time we use the Algorithm, we always want to take a look at the passes.  We're already using JSON heavily to store elements of the configuration, so just leaning more into this strategy for storing the passes.

<--------------------- REMOVE THE LINES BELOW BEFORE MERGING --------------------->

## Checklist
Please review and complete the following checklist before submitting your pull request:

- [x] I have ensured that the pull request is of a manageable size, allowing it to be reviewed within a single session.
- [x] I have reviewed my changes to ensure they are clear, concise, and well-documented.
- [x] I have updated the documentation, if applicable.
- [x] I have added or updated test cases to cover my changes, if applicable.
- [x] I have minimized the number of reviewers to include only those essential for the review.

## Checklist for Reviewers
Please review and complete the following checklist during the review process:

- [ ] The code follows best practices and conventions.
- [ ] The changes implement the desired functionality or fix the reported issue.
[...]

PR #253, opened 2025-03-17
allowing multiple race values

### LINKED ISSUE #250: Capture additional race info
## Summary
The [race is captured as a singular data element](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/schemas/pii.py#L260), however its possible that more than one race exists on an incoming document ([as is the case in FHIR](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/hl7/fhir.py#L62-L64)).  We need to allow for multiple race values in a submission and update our FHIR parsing logic to capture all of them, not just the last one.

## Impact
Losing race information in our records can negatively impact patient to patient comparisons.

## Expected behavior
- Allow users to submit more than one race value in our PIIRecord
- Yield all the possible RACE values in the feature_iter
- Capture all the race values submitted in the FHIR document, not just the last one

ericbuckley:
## Description
Allowing for more than 1 race value to be specified per payload.  Also fixing a bug in the FHIR parser that only accepted the last race specified, now all race values are processed.

## Related Issues
closes #250 

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -139,6 +139,29 @@ class Race(enum.Enum):
     ASKED_UNKNOWN = "ASKED_UNKNOWN"
     UNKNOWN = "UNKNOWN"
 
+    @classmethod
+    @functools.cache
+    def parse(cls, value: str) -> "Race":
+        """
+        Parse a race string into a Race enum.
+        """
+        # Create a list of string/race mappings, this is intentionally ordered
+        # to ensure we test for the substrings in the correct order
+        mapping = [
+            (["american indian", "alaska native"], cls.AMERICAN_INDIAN),
+            (["asian"], cls.ASIAN),
+            (["black", "african american"], cls.BLACK),
+            (["white"], cls.WHITE),
+            (["hawaiian", "pacific islander"], cls.HAWAIIAN),
+            (["asked unknown", "asked but unknown"], cls.ASKED_UNKNOWN),
+            (["unknown"], cls.UNKNOWN),
+        ]
+        val = value.lower().strip()
+        for substrings, race in mapping:
+            if any(substring in val for substring in substrings):
+                return race
+        return cls.OTHER
+
     def __str__(self):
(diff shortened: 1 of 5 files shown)


src/recordlinker/schemas/pii.py
```diff
@@ -399,8 +406,9 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
                 if name.family:
                     yield normalize_text(name.family)
         elif attribute == FeatureAttribute.RACE:
-            if self.race and self.race not in [Race.UNKNOWN, Race.ASKED_UNKNOWN]:
-                yield str(self.race)
+            for race in self.race:
+                if race and race not in [Race.UNKNOWN, Race.ASKED_UNKNOWN]:
+                    yield str(race)
```
m-goggins, line 411: This change has me second guessing whether we should allow race to be included in `FeatureAttribute`. We don't have log odds for race so we can't use it for blocking or evaluation (assuming we move forward with the log odds window approach and moving blocking to be reliant on log odds as well). 

This is likely beyond the scope of this ticket but it seems that `FeatureAttribute` should be based on the data that is available from the algorithm. If a user has log odds for race, then `FeatureAttribute.RACE` is valid. Thoughts on making `FeatureAttribute` more reliant on the supplied algorithm specification instead of an enum?

ericbuckley: Oh, that's in the other PR!  Check out this feature https://github.com/CDCgov/RecordLinker/pull/249/files#diff-f8b75e7f2665e1eea4eee28b690b98fa9fd3b2687327a847c80df04145ecb069R229

bamader (approved):
This looks good, since we'll be getting the change to FeatureAttribute as part of the kwargs PR.

src/recordlinker/schemas/pii.py
m-goggins: Oh nice, will review that separately!

Issue #223, opened 2025-02-21
remove kwargs from AlgorithmPass


ericbuckley:
## Summary
Remove the `AlgorithmPass.kwargs` field and replace it with specific fields for fuzzy matching and log odds parameters.

## Acceptance Criteria
- [ ] Add `log_odds` to Algorithm in schema/algorithm.py and models/algorithm.py
- [ ] Add `fuzzy_match_threshold` to Algorithm in schema/algorithm.py and models/algorithm.py
- [ ] Add `fuzzy_match_measure` to Algorithm in schema/algorithm.py and models/algorithm.py
- [ ] Add optional `fuzzy_match_threshold` to AlgorithmPass.evaluators in schema/algorithm.py and models/algorithm.py
- [ ] Add optional `fuzzy_match_measure` to AlgorithmPass.evaluators in schema/algorithm.py and models/algorithm.py
- [ ] Add `true_match_threshold` to AlgorithmPass in schema/algorithm.py and models/algorithm.py
- [x] Update logic in matchers.py to use the new values created above
- [x] Remove `kwargs` from AlgorithmPass in schema/algorithm.py and models/algorithm.py
- [ ] Add validation that log_odds are valid features and values are greater than 0
- [ ] Add validation to evaluators to verify that log_odds feature is available before use
- [ ] Add validation to blocking keys to verify that log_odds feature is available before use
- [] Add validation that fuzzy_match_threshold is between 0 and 1
- [] Add validation that fuzzy_match_measure [is valid](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/linking/matchers.py#L20)
- [x] All "defaults" options are optional and come with system defaults
- [ ] Add validation that true_match_threshold is less than the sum of the evaluator field log odds
- [x] Add good descriptive docstrings to describe the new values in the API docs
- [x] Remove kwargs information from site/reference.md
- [ ] Test cases for all of the above

## Details / Tasks
Use the snippet below as a basis for the additions that should be made to account for the deletion of the `kwargs` attribute.  Note, specific require Algorithm/AlgorithmPass attributes have been omitted and replaced with `...` as to highlight the additions that are being made.
```json
    {
        ...
        "log_odds": [
            {"feature": "ADDRESS", "value": 8.4},
            ...
        ],
        "defaults": {
          "fuzzy_match_threshold": 0.9,
          "fuzzy_match_measure": "JaroWinkler"
        },
        "passes": [
            {
                ...
[...]

PR #254, opened 2025-03-18
missing blocking keys

### LINKED ISSUE #230: improve blocking with missing payload keys
## Summary
Update the blocking query to skip optionally skip over missing blocking keys.

## Acceptance Criteria
- [ ] A reimplementation of get_block_data
- [ ] Add documentation to the site/design.md about how blocking uses log odds
- [ ] Add new algorithm configuration variable, "defaults/compare_minimum_percentage" with a default of 0.7.

## Details / Tasks
Modify the get_block_data function to optionally skip over the blocking key if its values are missing.  If we total all the log odds scores for the blocking keys in a pass, we must have at least the `default/compare_minimum_percentage` keys available to block on, otherwise we should just abort the query and return the empty dataset as the query will be too large.

## Background / Context
When an incoming payload has missing data, specifically data elements missing in blocking keys, we only return records that also have missing data in those fields.  This rewards returning incomplete records and penalizing those records that have values in those fields.  In some instances, it would be advantageous to also get back records with data in those fields.  On the flip side, we need to balance this, as if we skip too many blocking keys because of missing data we run the risk of returning too many records to analyze.

Our algorithms already store log odds values, which give us some indication as to how valuable a match is in record linkage.  We can reuse those values in blocking to make decisions on whether blocking keys can be skipped.  If too many values are skipped, we should just return the empty set and skip the pass altogether.

## Related Issues/PRs
#223

ericbuckley:
## Description
Changed the get blocking data call to allow for some missing blocking values.  If too many values are missing from the query (indicated by checking log odds values), then the blocking pass will be skipped.

## Related Issues
closes #230

## Additional Notes
A couple of things to note on the implementation.

1. The naming conventions for the new "compare_minimum_percentage" value does not match the issue.  This was done because the issue was written with the assumption that #223 would be completed first, which it is not.  The plan is to make changes to #223 to adjust the location of this new parameter, but for now we can just continue to use `kwargs`.
2. `get_block_data` was converted into a class, `GetBlockData`.  The [Improve Evaluation of Missing Values](https://github.com/CDCgov/RecordLinker/issues#workspaces/dibbs-record-linker-678acc70237f5000299a9d71/epics/Z2lkOi8vcmFwdG9yL1plbmh1YkVwaWMvMjI2OTc0) epic has introduced extra conditions for retrieving blocking data.  Checking for those extra conditions requires reusing variables, or at least benefits from reuse, thus storing some sort of state between the different functions has an advantage.  There are some alternatives, that are worth considering, that I experimented with before landing on this solution.   I think all three implementations can do a good job of encapsulating the logic and optimizing the looping constructs for efficient evaluation.  If anyone thinks another implementation would read better, please voice that.
    - A long function with all the logic encapsulated
    - A long function with nested functions

## Checklist

## Checklist for Reviewers


src/recordlinker/database/mpi_service.py
```diff
@@ -17,118 +18,153 @@
...
+                expression.and_(
+                    models.Patient.id == alias.patient_id,
+                    alias.blockingkey == key.id,
+                    alias.value.in_(self.blocking_values[key]),
+                ),
+            )
+
+        if self._check_skip_conditions():
+            # Too many missing blocking values
+            return []
```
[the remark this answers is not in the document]
m-goggins: Hm I was suggesting that would re-order when we call `_check_skip-conditions` to check before we get to `self.blocking_values[key] = [v for v in record.blocking_keys(key)]`. 

For example, let's say we have an incoming record that is missing BIRTHDATE but has SEX and IDENTIFIER. In the first pass, as it currently works, it checks that the record has BIRTHDATE, then IDENTIFIER, and then SEX before deciding whether to skip blocking. In my proposal (which would need to be fleshed out still), we would check if the record has a BIRTHDATE and also check whether we should proceed with checking the remaining block values. Since a record without BIRTHDATE does not have enough log odds points to proceed with only SEX and IDENTIFIER, we know that we don't have to even check if the record has those values with `self.blocking_values[key] = [v for v in record.blocking_keys(key)]`

tl;dr: If we already know the total logs odds for a pass and we check more frequently if a record meets the `_check_skip-conditions`, I think we could reduce the number of times we call `self.blocking_values[key] = [v for v in record.blocking_keys(key)]`.

Example:
1. calculate total log odds for pass 1: BIRTHDATE (10) + SEX (0.75) + IDENTIFIER (0.3) = ~11
2. For the 1st blocking value (which has the most value), check if the incoming record is missing. Let's say it is missing BIRTHDATE.
3. Check whether we have enough points to proceed: Since BIRTHDATE was missing we only have 1/11 points available (at best), which is below 0.7. We should not check whether the incoming record has values for SEX or IDENTIFIER.
4. return [] since the incoming record is missing too many values.

Does that make sense?

src/recordlinker/assets/initial_algorithms.json
```diff
@@ -47,7 +47,9 @@
                         "SEX": 0.7510419059643679,
                         "STATE": 0.022376768992488694,
                         "ZIP": 4.975031471124867
-                    }
+                    },
+                    "compare_minimum_percentage": 0.7
```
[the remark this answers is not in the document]
bamader: Hmm, I think I can see the value of having different blocking and evaluation proportions for maximum allowed missingness, if that's something we want to do. I could especially see the value of this to set a higher proportion of required points for blocking, like 0.7 present, to keep candidate sets small. But I don't have too strong a preference on that front, especially since the DIBBs algorithm won't leverage too much either way. However, what I do like and think we should keep are the names @m-goggins and I came up with for handling missingness during evaluation, 'max_allowed_missingness_proportion' and 'missing_field_points_proportion'. We like the standardization on proportion denoting a 0 to 1 range and I personally feel the names are much clearer and descriptive this way. So I'd say if we want to keep a separate parameter for blocking, we should give commensurate names that follow the same convention, so something like 'max_blocking_missingness_allowed_proportion'. More verbose maybe, but ultra clear.

src/recordlinker/routes/algorithm_router.py
```diff
@@ -58,11 +59,15 @@ def create_algorithm(
...
     """
     try:
-        service.load_algorithm(session, data)
+        service.load_algorithm(session, data, commit=True)
     except ValueError as exc:
         raise fastapi.HTTPException(
             status_code=fastapi.status.HTTP_422_UNPROCESSABLE_ENTITY, detail=str(exc)
         )
+    except sqlexc.IntegrityError as exc:
+        raise fastapi.HTTPException(
```
ericbuckley, line 68: @m-goggins @bamader this is unrelated to the GH issue, but a bug I found while testing yesterday.  If a user tries to create an algorithm with a non-unique label, a server error is generated rather than a 422.  I hate seeing 500s in the code, so I try to fix those as fast as possible, and thus put it in here.

I added a test case for this in test_algorithm_router.py and verified locally that this block was working as expected. I'm not sure why codedov is saying otherwise.

src/recordlinker/database/mpi_service.py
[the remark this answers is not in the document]
ericbuckley: It does, I'll give it a shot.

src/recordlinker/assets/initial_algorithms.json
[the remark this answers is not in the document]
ericbuckley: If there's not a strong reason to, I think we should reuse the existing variable.  Configuration has a lot of parameters already, and I don't want to add another if it's not necessary.  I'll just reuse `max_allowed_missingness_proportion` if that works.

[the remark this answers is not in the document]
ericbuckley: renamed.  https://github.com/CDCgov/RecordLinker/pull/254/commits/87bffa7781b39af7d0c204551ff09397d396ae04

src/recordlinker/database/mpi_service.py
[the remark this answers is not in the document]
m-goggins: Yes, exactly!

src/recordlinker/database/mpi_service.py
```diff
@@ -17,118 +18,153 @@
...
+                    alias.value.in_(self.blocking_values[key]),
+                ),
+            )
+
+        # Using the subquery of unique Patient IDs, select all the Patients
+        expr = expression.select(models.Patient).where(models.Patient.person_id.in_(base))
+        # Execute the query and collect all the Patients in matching Person clusters
+        patients: typing.Sequence[models.Patient] = session.execute(expr).scalars().all()
+        # Remove any Patient records that have incorrect blocking value matches
+        return [p for p in patients if self._filter_incorrect_match(p)]
```
bamader, line 167: Small thing, but the language consistency between `_filter_incorrect_match()` and `_check_skip_conditions()` seems off to me. The former returns `True` when something does _not_ get ruled out, but the latter returns `True` when a pass _does_ get ruled out. I think this makes things a bit less intuitive to read. Can we standardize the naming between those functions so that for both of them, returning `True` means the same thing (likely that something is kept)?

Personally, I also like naming boolean test / filter functions with conventions like `_should_[...]` or to-be verbs like `X_is_Y` / `X_has_Y` to indicate that the function is actively performing a membership test, but that one's more up to dealer's choice whereas I think the standardization is important for general readability. I'd propose names like `_should_continue_blocking()` and `_patient_agrees_on_blocking()`, but just as starter suggestions.

src/recordlinker/database/mpi_service.py
```diff
@@ -17,118 +18,153 @@
...
-        for bk, allowed_vals in blocking_vals_in_incoming.items():
-            # Compare incoming blocking value to what would be the blocking
-            # value of the mpi record to make sure we compare on e.g. same
-            # number of characters at beginning/end of string
-            mpi_vals = mpi_record.blocking_keys(getattr(models.BlockingKey, bk))
-
+LOGGER = logging.getLogger(__name__)
+
+
+class GetBlockData:
```
bamader, line 24: (Non-blocking) What's the thinking on making this a class versus a collection of functions? It kind of threw me for a loop when I first saw it because I didn't find the relationship between functions + their purposes and a class structure very intuitive (the class seems strictly functional-output oriented so I had a bit of a hurdle in reading it).

src/recordlinker/linking/link.py
```diff
@@ -165,7 +165,8 @@ def link_record_against_mpi(
             with TRACER.start_as_current_span("link.block"):
                 # get all candidate Patient records identified in blocking
                 # and the remaining Patient records in their Person clusters
-                pats = mpi_service.get_block_data(session, record, algorithm_pass)
+                blocker = mpi_service.GetBlockData()
```
bamader, line 168: Somewhat related to my question around why make it a class, but do we think this will have a meaningful impact on time and compute performance? It's not a big cost, but there is still a cost when instantiating a class with every call to what's going to be our most commonly invoked function. Do we get some benefit here I'm not seeing?

src/recordlinker/routes/algorithm_router.py
bamader: Oh wonky. Thanks for letting us know! I was gonna flag it as needing coverage but if you're good locally then yeah I don't know what's up with codecov. Also good catch on that error--I hate debugging and seeing 500s so anything to clean them up is a good change in my eyes.

bamader (commented):
Looks good overall, some questions around class decisions and naming conventions, but I think this is basically there

src/recordlinker/database/mpi_service.py
ericbuckley: I added some notes to the PR description, does this help clarify?

get_block_data was converted into a class, GetBlockData. The [Improve Evaluation of Missing Values](https://github.com/CDCgov/RecordLinker/issues#workspaces/dibbs-record-linker-678acc70237f5000299a9d71/epics/Z2lkOi8vcmFwdG9yL1plbmh1YkVwaWMvMjI2OTc0) epic has introduced extra conditions for retrieving blocking data. Checking for those extra conditions requires reusing variables, or at least benefits from reuse, thus storing some sort of state between the different functions has an advantage. There are some alternatives, that are worth considering, that I experimented with before landing on this solution. I think all three implementations can do a good job of encapsulating the logic and optimizing the looping constructs for efficient evaluation. If anyone thinks another implementation would read better, please voice that.
- A long function with all the logic encapsulated
- A long function with nested functions

src/recordlinker/database/mpi_service.py
ericbuckley: Consistency's good, thanks for calling that out.  I do like the `_filter_incorrect_match` name as it adheres to the standard [filter syntax](https://docs.python.org/3/library/functions.html#filter) and I think the name indicates that.  I am however by no means tied to the `_check_skip_conditions` name, so changing to `_should_continue_blocking` and reversing the bools works for me.

bamader: That sounds good. Aligning with other filter functions is a good priority. Thanks for the flexibility on the other name!

src/recordlinker/linking/link.py
ericbuckley: `GetBlockData` inherits from `object`, and is whats commonly called a simple class.  The price to instantiate is extremely low (probably a couple dozen microseconds on modern hardware).  I just ran a test, and its maybe 50% more expensive than creating the 3 variables in _reset.

```
❯ python -c "import timeit; print(timeit.timeit('for _ in range(1_000_000): blocker: object = object()', number=1))"
0.03495995904086158
❯ python -c "import timeit; print(timeit.timeit('for _ in range(1_000_000): total_odds: float = 0.0; missing_odds: float = 0.0; blocking_values: dict[str, list[str]] = {}', number=1))"
0.022246917011216283
```

I can certainly move the variable creation to outside the for loop, but right now our default passes is 2, so its not a big savings.  But there's also no reason why it shouldn't be declared higher up, so I'll do that.

src/recordlinker/assets/initial_algorithms.json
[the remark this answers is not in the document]
bamader: Oh, missed commenting on this one, but yeah I don't really see a compelling reason to preserve both variables, so thanks for the rename and we can consider this handled.

src/recordlinker/linking/link.py
bamader: Gotcha. This for sure isn't a meaningful performance hit, and you've already explained the motivation for re-using state in the class. Definitely don't feel like you have to move any variables around, your test shows this isn't going to make any real difference to our users. Write it up however makes sense to you! Just wanted to ask the question since the `link_` function will be called more than any other 👍

bamader (approved):
I think all discussions have been handled, so thanks for some of the naming/convention changes and the explanations of thought processes.

src/recordlinker/database/mpi_service.py
ericbuckley: Hey, it's a great thought, and with a class it's easy to encapsulate the logic so it's obvious to the reader.  The more I look at the current implementation, the `_reset` method gives off a smell that state isn't really necessary.  When implementing the current version, I was thinking a lot of parameters would have to be passed to private functions, which really isn't the case.  And the `_filter_incorrect_match` couldn't be a real filter function without a closure and/or state.  But that really isn't important and I didn't even use the filter builtin, I went with a comprehension so its moot.  Let me know what you think of this new implementation.  https://github.com/CDCgov/RecordLinker/pull/254/commits/1a354fe2ef443b797d653c8866e03d4945f761ad

m-goggins (approved):

PR #259, opened 2025-03-20
Handle missing fields during evaluation

### LINKED ISSUE #235: Improve evaluation of missing data fields
## Summary
Evaluating two records on a feature, where one of those records doesn't have a value, penalizes the comparison too much.  For example, if two birthdates don't match (1980-01-01 and 1990-06-05) that should result in a worse score than a record with 1980-01-01 and a record that is missing a birthdate.  Right now, the former gets a better score.

## Acceptance Criteria
- [ ] Add a new algorithm config value of "defaults/compare_missing_percentage", default to 0.5
- [ ] Update the link.compare method
- [ ] Update the matchers.compare_* signatures to indicate when data is missing
- [ ] New test cases for the compare method

## Details / Tasks
The compare method needs to use the new "defaults/compare_missing_percentage" and the "defaults/compare_minimum_percentage" when calculating the value.  When comparing the incoming record to an existing one, `compare_missing_percentage` of the log odds value should be rewarded for the feature comparison if either value is missing.

When comparing two records, if more than `compare_minimum_percentage` is missing from the comparison, the evaluation as a whole should result in a 0 score.

## Dependencies
#230

bamader:
## Description
This PR adds more robust handling of comparisons between records in which one or more fields are missing data. If only a small amount of data is missing (fewer log-odds-worth than a specified percentage of the total possible), then missing comparisons are awarded partial weight to allow for data recovery and quality matches. However, if too much data is missing, the candidacy is automatically rejected.

## Related Issues
Closes #235 

## Additional Notes
Just changed some variable names from the ticket for clarity

## Checklist

## Checklist for Reviewers

### DIFF
--- a/docs/site/reference.md
+++ b/docs/site/reference.md
@@ -174,11 +174,11 @@ In that case we'd want to evaluate "123 Main St" against both "123 Main Street" 
     based on training. Example: `{"kwargs": {"thresholds": {"FIRST_NAME": 0.8}, "log_odds": {"FIRST_NAME": 6.8}}}`
 
 One important caveat for both of these Feature Functions is how they handle patient
-records with missing information in one or more fields. RecordLinker has the potential
-to allow records that may be missing some data to "recover" this information by matching
-with other records for which that data is present. In order to enable this possibility,
-and to avoid overly penalizing records which may be strong matches but simply have data
-omitted due to collection methods, both of these feature functions include a partial
-log-odds weighting. If one or more records being compared is missing data for a field,
-each of the above functions returns exactly half the log-odds weight for the field
-overall, along with a boolean flag indicating that data was missing during comparison.
+records with missing information in one or more fields.  RecordLinker provides the option 
+to match records that are missing some data, e.g., Field X, with other records for which that 
+data (Field X) is present. In order to enable this possibility, and to avoid overly penalizing 
+records which may be strong matches but simply have data omitted due to collection 
+methods, both of these feature functions include a partial log-odds weighting. If one or more 
+records being compared is missing data for a field, each of the above functions returns exactly 
+half the log-odds weight for the field overall, along with a boolean flag indicating that data was 
+missing during comparison.
--- a/docs/site/reference.md
+++ b/docs/site/reference.md
@@ -172,3 +172,13 @@ In that case we'd want to evaluate "123 Main St" against both "123 Main Street" 
     that feature is in determining whether two patient records are a true match, as opposed to a match
     by random chance). Use the kwargs parameter to specify the fuzzy match threshold and log-odds ratio
     based on training. Example: `{"kwargs": {"thresholds": {"FIRST_NAME": 0.8}, "log_odds": {"FIRST_NAME": 6.8}}}`
+
+One important caveat for both of these Feature Functions is how they handle patient
+records with missing information in one or more fields. RecordLinker has the potential
+to allow records that may be missing some data to "recover" this information by matching
+with other records for which that data is present. In order to enable this possibility,
+and to avoid overly penalizing records which may be strong matches but simply have data
+omitted due to collection methods, both of these feature functions include a partial
+log-odds weighting. If one or more records being compared is missing data for a field,
+each of the above functions returns exactly half the log-odds weight for the field
+overall, along with a boolean flag indicating that data was missing during comparison.
--- a/src/recordlinker/assets/initial_algorithms.json
+++ b/src/recordlinker/assets/initial_algorithms.json
@@ -8,8 +8,8 @@
             0.75,
             0.9
         ],
-        "max_missing_field_proportion": 0.5,
-        "missing_field_compare_fraction": 0.5,
+        "max_missing_allowed_proportion": 0.5,
+        "missing_field_points_proportion": 0.5,
         "passes": [
             {
                 "blocking_keys": [
--- a/src/recordlinker/assets/initial_algorithms.json
+++ b/src/recordlinker/assets/initial_algorithms.json
@@ -1,97 +1,99 @@
 [
-    {
-        "label": "dibbs-default",
-        "description": "The core DIBBs Log-Odds Algorithm. This default, recommended algorithm uses statistical correction to adjust the links between incoming records and previously processed patients (it does so by taking advantage of the fact that some fields are more informative than others—e.g., two records matching on MRN is stronger evidence that they should be linked than if the records matched on zip code). It can be used if additional granularity in matching links is desired. However, while the DIBBs Log-Odds Algorithm can create higher-quality links, it is dependent on statistical updating and pre-calculated population analysis, which requires some work on the part of the user. For those cases where additional precision or stronger matching criteria are required, the Log-Odds algorithm is detailed below.",
-        "is_default": true,
-        "include_multiple_matches": true,
-        "belongingness_ratio": [
-            0.75,
-            0.9
+  {
+    "label": "dibbs-default",
+    "description": "The core DIBBs Log-Odds Algorithm. This default, recommended algorithm uses statistical correction to adjust the links between incoming records and previously processed patients (it does so by taking advantage of the fact that some fields are more informative than others—e.g., two records matching on MRN is stronger evidence that they should be linked than if the records matched on zip code). It can be used if additional granularity in matching links is desired. However, while the DIBBs Log-Odds Algorithm can create higher-quality links, it is dependent on statistical updating and pre-calculated population analysis, which requires some work on the part of the user. For those cases where additional precision or stronger matching criteria are required, the Log-Odds algorithm is detailed below.",
+    "is_default": true,
+    "include_multiple_matches": true,
+    "belongingness_ratio": [
+      0.75,
+      0.9
+    ],
+    "max_missing_field_proportion": 0.5,
+    "missing_field_compare_fraction": 0.5,
+    "passes": [
+      {
+        "blocking_keys": [
+          "BIRTHDATE",
+          "IDENTIFIER",
+          "SEX"
         ],
-        "passes": [
-            {
-                "blocking_keys": [
-                    "BIRTHDATE",
-                    "IDENTIFIER",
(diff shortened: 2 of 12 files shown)


src/recordlinker/linking/link.py
```diff
@@ -47,18 +55,32 @@ def compare(
...
+        
+        # Evaluate the comparison function, track missingness, and append the 
+        # score component to the list
+        result: tuple[float, bool] = e.func(
+            record, patient, feature, missing_points_proportion, **kwargs
+        )  # type: ignore
+        if result[1]:
+            # If the field was missing, the log-odds weight was multiplied, so dividing
+            # here reverses the change without needing to overly pass the odds dict
+            missing_field_weights += result[0] / missing_points_proportion
```
m-goggins, line 75: Is `missing_points_proportion` the same as `missing_field_compare_fraction` in models/algorithm.py? If they are the same, I'm worried that setting the `missing_points_proportion` to 0.0 would break here. 

Can you add a test to address this scenario?

m-goggins (changes requested):
Overall, this is looking solid. Most of my comments were about making sure we use consistent naming from model/schema definitions, `link.py, and `matchers.py` to make it easier to understand how all the pieces fit together. 

I do worry about recalculating the `missing_field_weights` after evaluating in `compare` if a user can set `missing_points_proportion` to 0.0. Ultimately, I think we should let users set that field to be 0.0 if they want (even though we know this punishes missing values too much) so we probably need to account for that in our function and tests.

docs/site/reference.md
```diff
@@ -172,3 +172,13 @@ In that case we'd want to evaluate "123 Main St" against both "123 Main Street"
...
+
+One important caveat for both of these Feature Functions is how they handle patient
+records with missing information in one or more fields. RecordLinker has the potential
+to allow records that may be missing some data to "recover" this information by matching
+with other records for which that data is present. In order to enable this possibility,
+and to avoid overly penalizing records which may be strong matches but simply have data
+omitted due to collection methods, both of these feature functions include a partial
+log-odds weighting. If one or more records being compared is missing data for a field,
+each of the above functions returns exactly half the log-odds weight for the field
+overall, along with a boolean flag indicating that data was missing during comparison.
```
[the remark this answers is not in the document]
bamader: That's completely fair and a good point. Dan and I started using that language way back with VA and I've never been able to drop it. Your version is much clearer.

src/recordlinker/assets/initial_algorithms.json
```diff
@@ -8,6 +8,8 @@
             0.75,
             0.9
         ],
+        "max_missing_field_proportion": 0.5,
```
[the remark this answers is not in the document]
bamader: Yeah I agree completely. We have the `site/reference.md` and a `site/configuration.md`, but the former is fairly barebones and the latter pertains to DB setup. I think we should write some good, user-friendly documentation about what all our values mean and how to structure things, probably with a pass from Ryan (or Matt Goldberg) to make sure it's readable.

src/recordlinker/linking/link.py
```diff
@@ -35,10 +35,18 @@ class LinkResult:
...
+    record: schemas.PIIRecord,
+    patient: models.Patient,
+    max_log_odds_points: float,
+    max_allowed_missingness_proportion: float,
+    missing_points_proportion: float,
+    algorithm_pass: models.AlgorithmPass
 ) -> bool:
     """
-    Compare the incoming record to the linked patient
+    Compare the incoming record to the linked patient and return the calculated
```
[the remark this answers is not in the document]
bamader: For sure. Maybe we want to revisit the idea of a norm around full docstrings, like we used to have. I've begun adding them in some places but if we want autogenerated docs to be maximally effective I don't see a reason that every function (that isn't an internal `_helper()` style) shouldn't have docstrings.

src/recordlinker/linking/link.py
bamader: Should be handled now that we're passing in the log odds dict directly 👍

bamader:
Names for variables should be fully standardized, and the division by 0 case should be avoided now that we directly pass in log-odds.

bamader:
@m-goggins No problem at all! I'll actually do you one better--your comment about testing 0's in both fields made me think of an "edge case" where the user sets `missing_fields_proportion` to be much larger than `missingness_allowed` (I outlined a scenario in the tests where this might happen). The concept is a bit the same as some of the other tests, but I thought it was worth calling out that we manage results from both the perspective of missing field points and missingness overall, regardless of which one is a larger proportion.

m-goggins (approved):
Awesome and good thought about the additional edge case!

src/recordlinker/linking/matchers.py
```diff
@@ -116,68 +116,104 @@ def rule_probabilistic_match(feature_comparisons: list[float], **kwargs: typing.
...
     """
     log_odds = kwargs.get("log_odds", {}).get(str(key.attribute))
     if log_odds is None:
         raise ValueError(f"Log odds not found for feature {key}")
 
+    # Return early if a field is missing, and log that was the case
+    incoming_record_fields = list(patient.record.feature_iter(key))
+    mpi_record_fields = list(record.feature_iter(key))
+    if len(incoming_record_fields) == 0 or len(mpi_record_fields) == 0:
+        return (missing_field_points_proportion * log_odds, True)
```
ericbuckley, line 155: I know this is merged, but I have an upcoming refactor on #249 to rebase these changes.

If the signature was changed to return a bool if missing data is encountered, is it necessary to calculate the missing field points proportion in the compare_* methods?  It seems like this will be duplicated logic across all compare_* methods, and if we just do this on the link.compare function we can dry it up.  Thoughts?

bamader: If I understand the comment, you're saying line 155 contains a calculation that happens in two different functions (`compare_exact` and `compare_fuzzy`). In my eyes, that calculation belongs inside both compare functions and is not an instance of unnecessary code duplication (it's also just two functions). The idea of the scope of a `compare` function is to tell the caller how many points a candidate match should earn for a particular field. That means it is responsible for applying all properties and parameters needed to determine that value. This is why `compare_fuzzy` for example applies thresholding--because it's part of determining how much value a field comparison earns. While it's possible for the calling `compare` loop to do this by looking up the feature in question in the log-odds dictionary, I believe this is a) less clear and intuitive, b) defeats the abstraction and scope of self-contained compare functions (which would no longer be independent producers of the field value), and c) introduces unnecessary case-by-case bases in scoring, with results sometimes sometimes being produced in the `compare_*` and sometimes being in `compare`. In my opinion, these `compare_*` functions are correctly written as-is to make the logic the clearest and allow each function to be modular.

ericbuckley: I totally get that, and re-reading my previous comment, "dry it up" was the wrong line of thought to my concern.

When I think about the two new configuration values, "max_missing_allowed_proportion" and "missing_field_points_proption", they feel inherently linked. From a readability perspective, it might make sense to handle their logic together in the `link.compare` function. This way, the relationship between them is more apparent to the next developer reading the code. Maybe I’m balancing readability against functional correctness?

No need to act on this right now, I'm going to experiment with both approaches in #249.  I guess just keep this in mind for the conversation we have on that PR.

bamader: Ah okay, thanks for clarifying. I see a little bit better what you're saying now.

Interestingly, I actually have the opposite view. To me, the values would be linked if setting one somehow influenced or determined a range of values to set the other (they might be correlated in that way). But I actually found in thinking through implementation, I believe there to be reasonable, plausible scenarios for which users might set them at completely opposite ends of the spectrum. For example, in DIBBs default, we have only a small number of evaluation fields, each with very large log odds weight. In this context, we likely want to set our `max_missing_allowed_proportion` somewhere low, so that missing one field means the record isn't likely to make a comparison match. But the reverse situation is also conceivable (and we added a test case for it)--a user might have an algorithm set up in which they want to compare on many different fields, but none of their individual log-odds values are massively dominant over all others. In this use context, I could see justifications for having a high missingness allowed and low points for missing fields (if I knew my data quality was going to be really messy, which we know empirically is the case for streams like eLR), as well as a justification for having a low missingness allowed but a high points for missing fields (where I want too much missing data to disqualify a match but only a little bit to not be a hindrance if I care about finding true matches).

I bring this up because I think it underscores another point I alluded to in my comment, which is the separation between a single field comparison function and a looping comparison function. I think our code benefits a lot from the modularity of having a single function (or function type in `compare_*`) be solely responsible for deciding how many points a field comparison is worth. If I'm a user trying to make sense of where that value comes from, I have to look in a single place in the code, in the location where all the other field-wise parameters are accounted for. It makes intuitive sense to me. It also makes sense from the other side, thinking about the responsibility of the looping `compare` function. This is more like an orchestrator; it's not a function that gets in the weeds with arbitrating results on individual field comparisons. It's a function that tells subordinates it calls "you tell me what your field status is, and I'll figure out the best way to use that information." There's a cleanness there, in my opinion, a self-containment that actually makes the code more readable by not grouping the parameters.

A related example is something like thresholding. Right now, the `compare_fuzzy` function does its own thresholding to 0 if the jaro winkler similarity is below the configuration-specified value. That makes sense to me for the same reasons as above, namely that one function is presenting its caller with the final verdict on how many points a field is worth (and the decision to drop the MPI match candidate from consideration doesn't actually change how many points that field is worth, that's always the same number for a given algorithm config). If we pulled that up into `compare`, we reduce modularity and in my opinion sacrifice readability. Now, sometimes the results from the `compare_fuzzy` function can be trusted as final, if a field score's jaro winkler is above the threshold; but other times, it can't be, if the jaro winkler similarity is low. And the fact that we don't know this when we evaluate means the functions are now no longer independent, and in my opinion this also makes it harder for a user or future engineer to understand and debug. Curious to hear @m-goggins thoughts on this.

I think what I'm surfacing for myself is that I would rather see us go slowly on reorganizing kwargs and the algorithm config because it might have readability and intuition consequences on other parts of the code.

Issue #244, opened 2025-03-10
Similarity comparisons should ignore whitespace


m-goggins:

m-goggins:
Duplicate of #243

PR #261, opened 2025-03-21
Add caching of RecordLinker image to smoke tests

### LINKED ISSUE #224: increase docker caching in workflows
## Summary
On Apr 1, Docker will implement a change to reduce [unauthenticated requests to just 10 pulls an hour](https://docs.docker.com/docker-hub/usage/).  We need to be sure that our workflows are caching docker images as much as possible to reduce/prevent our chances of running into 429 failures.

## Acceptance Criteria
- [ ] Adjustments to the github action workflows to improve docker cache hits

## Details / Tasks
Areas to consider for improvements
- The smoke test job uses the `docker build ...` command to create an image to run.  This likely hits Docker hub to retrieve the python:3.12-slim image every time.  Consider using [docker/build-push-action](https://github.com/docker/build-push-action) to make use of caching.
- Our testing workflows depend on database services that use Docker hub images.  It's likely that Github is caching some of these images, but in the case of an image like `....-latest` it's probably having to check Docker hub every time to see if the image has changed.  Maybe pinning these services to tags that don't change will prevent that (eg postgres:17 -> postgres:17.4-bookwarm)?  Another alternative is to use a github mirror, that is copy the images we want into a repo on ghrc.io/cdcgov/recordlinkage, then have our workflow services reference those images.
- Also, as part of an optimization, investigate whether the buildx step is necessary in the docs job or the release workflow.

m-goggins:
## Description
This PR addresses half of the caching issue outlined in ticket 224: ensuring that we use the cached the RecordLinker image when we run smoke tests to avoid hitting the limit on requests to DockerHub. 

After much tinkering, I added a `build-push-run` step to our smoke tests so now when you run them you should see that we use the image from ghcr for each smoke test instead of hitting `docker.io/library/python:3.12-slim` to build the service in each run.

## Related Issues
Fixes #224 

## Additional Notes
Getting the services to run from images in the GHCR has proven to be much more difficult, which is why I opted to split up the work on #244. At least we will have cut down the number of requests for RecordLinker by the 4/1 deadline.

## Checklist

## Checklist for Reviewers

### DIFF
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -51,7 +51,6 @@ jobs:
     - name: Build and push RecordLinker image
       uses: docker/build-push-action@v3
       with:
-          context: ./recordlinker
           push: true
           tags: |
             ghcr.io/${{ env.REPO }}/rl-service:latest 
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -40,7 +40,7 @@ jobs:
     steps:
     - name: Check out repository
       uses: actions/checkout@v4
-      
+
     - name: Convert repository name to lowercase
       run: echo "REPO=$(echo '${{ github.repository }}' | tr '[:upper:]' '[:lower:]')" >> $GITHUB_ENV
 
@@ -58,9 +58,9 @@ jobs:
           tags: |
             ghcr.io/${{ env.REPO }}/rl-service:latest 
 
-    - name: Pull cached RecordLinker image from GitHub Container Registry
-      run: |
-        docker pull ghcr.io/${{ env.REPO }}/rl-service:latest || echo "No cached image found, skipping pull."
+    # - name: Pull cached RecordLinker image from GitHub Container Registry
+    #   run: |
+    #     docker pull ghcr.io/${{ env.REPO }}/rl-service:latest || echo "No cached image found, skipping pull."
     
     - name: Start RecordLinker Service and Run Smoke Tests
       run: |
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -57,10 +57,6 @@ jobs:
           push: true
           tags: |
             ghcr.io/${{ env.REPO }}/rl-service:latest 
-
-    # - name: Pull cached RecordLinker image from GitHub Container Registry
-    #   run: |
-    #     docker pull ghcr.io/${{ env.REPO }}/rl-service:latest || echo "No cached image found, skipping pull."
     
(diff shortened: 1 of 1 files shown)


bamader (approved):
This looks super painful despite being so short 😞 . Thanks for working through these difficulties. Looking at the draft of the other work, splitting definitely seems like the right call to me.

.github/workflows/check_smoke_tests.yml
```diff
@@ -41,13 +41,23 @@ jobs:
...
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+      
+    - name: Build and push RecordLinker image
+      uses: docker/build-push-action@v3
+      with:
+          push: true
+          tags: |
+            ghcr.io/${{ env.REPO }}/rl-service:latest
```
ericbuckley, line 59: I think there is a potential for a conflict here if we have more than one open PR.  If both PR's are building, the last one win's out.

m-goggins: @ericbuckley I was thinking of changing the image name to be based on the pr like `rl-service:pr-$PR_NUMBER`, but want to make sure the image gets removed after the PR is merged. I could do it through a workflow or let a retention policy take care of it, but I don't know what our retention policy is and but don't seem to have the right permissions to do so. Do we have a retention policy set or should I remove the image with a workflow?

m-goggins: Will make any updates in the other PR to address this

ericbuckley: To my knowledge we don't have a retention policy set.  I'm not even sure how to do that on a ghcr.io repo, but I'm sure there is a way.  One thing that might be worth considering is using docker layer caching for this rather than an ghcr.io repo.  Take a look at this article https://medium.com/@depot.dev/how-to-use-docker-layer-caching-in-github-actions-bee1f7b19093

kylegalbraith: Author of ☝️ (co-founder of Depot, former PDX native), you can find a less annoying & updated version (Medium can be meh) of that post here: https://depot.dev/blog/docker-layer-caching-in-github-actions

Also, if you want to skip pushing to GHCR to use as the cache, you can specify `load: true`, and that _should_ just get the image back to you in its tagged form to then do a `docker run` on it. You could then do the `docker push` after your tests to put it in GHCR.

If you also want an open source sponsorship of Depot to swap `docker/build-push-action` for `depot/build-push-action`, also happy to set that up.

ericbuckley: Thanks @kylegalbraith I appreciate the guidance.

PR #262, opened 2025-03-21
Fix caching for services during smoke tests

### LINKED ISSUE #224: increase docker caching in workflows

m-goggins:
## Description
This is part 2 of ticket #224, which ensures that we are caching the RL image and the database images during the smoke tests. 

## Related Issues
Fixes #224


.github/workflows/check_smoke_tests.yml
```diff
@@ -7,132 +7,184 @@ on:
...
+              echo "Failed to pull source image: $SOURCE_IMAGE" >&2
+              exit 1
+            fi
+          else
+            echo "Image already exists in GHCR: $TARGET_IMAGE, skipping pull and push."
+          fi
+
   smoke_tests:
     runs-on: ubuntu-latest
+    needs: service-set-up
```
bamader, line 60: Ah crafty :neckbeard:

.github/workflows/check_smoke_tests.yml
```diff
@@ -7,132 +7,184 @@ on:
...
+              echo "RL Service did not become healthy in time."
+              exit 1
+            fi
+          done
+
+          # Run smoke tests and print the response
+          JSON_BODY_1='{"record": {"birth_date": "2013-11-07", "sex": "M", "identifiers":[{"value": "123456789", "type": "MR"}], "name":[{"family":"Shepard", "given":["John"]}]}}'
+          JSON_BODY_2='{"algorithm": "dibbs-default", "record": {"birth_date": "2000-12-06", "sex": "M", "identifiers":[{"value": "9876543210", "type": "MR"}], "name":[{"family":"Smith", "given":["William"]}]}}'
+
+          #basic tests
```
bamader, line 148: Completely unrelated thing that I just caught: this comment and the `enhanced tests` on line 166 are leftover from when we had DIBBs basic and DIBBs enhanced rather than just the one algorithm. Can we delete the comments just so there's no confusion on one set of tests using different functionality than the other? There's nothing different between them anymore.

bamader (approved):
This is the kind of PR where I understand each line individually but trying to piece the whole context together makes my brain hurt 🧠 . Great work on getting through a really hairy and tough to navigate set of code though!

.github/workflows/check_smoke_tests.yml
m-goggins: Oh good catch! Fixed

bamader (approved):
Looks good! Think I understand it 2% more the second time around 😅

.github/workflows/check_smoke_tests.yml
```diff
@@ -7,132 +7,182 @@ on:
     branches: [main]
 
 jobs:
+  service-set-up:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: true
+      matrix:
+        database: [postgres:17, mysql:8, mssql/server:2022-latest]
```
ericbuckley, line 16: One thing we should consider is pinning these databases.  Using "latest" specifically means that we will redownload an image whenever they publish anything new.  Using ":8" or ":17" means we will redownload an image whenever they make a minor release change.  Are those incremental changes necessary for our work?

.github/workflows/check_smoke_tests.yml
```diff
@@ -7,132 +7,182 @@ on:
...
+            SOURCE_IMAGE="mcr.microsoft.com/mssql/server:2022-latest"
+          else
+            TARGET_IMAGE="ghcr.io/${{ env.REPO }}/${{ matrix.database }}"
+            SOURCE_IMAGE="${{ matrix.database }}"
+          fi
+
+          # Check if the image exists in GHCR
+          if ! docker pull "$TARGET_IMAGE" >/dev/null; then
+            echo "Image not found in GHCR, pulling from source: $SOURCE_IMAGE..."
+            if docker pull "$SOURCE_IMAGE"; then
```
ericbuckley, line 46: Something to consider, we are also using database images in the check_unit_tests workflow.  I think with this strategy we'd have to copy these jobs there as well.  Is that right?

ericbuckley:
@m-goggins what do you think about the following strategy?

1) Create a new [scheduled workflow](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onschedule) that runs maybe once a month (open to suggestions on the timing).  This workflow does more or less what you have so far in the PR.  Mirrors the database images from dockerhub into ghcr.
2) Update the unit test and smoke test jobs to use the ghcr images rather than the dockerhub versions.

m-goggins:
@ericbuckley Yeah, I like that suggestion! Since we're using the database images across multiple tests, i think it make sense to have it more centralized as opposed to per workflow.

.github/workflows/check_smoke_tests.yml
m-goggins: Resolved by adding minor versions!

m-goggins:
@ericbuckley After trying out the `build-push-action` in the smoke tests for the RL image ([here](https://github.com/CDCgov/RecordLinker/pull/262/commits/e1232c9b5316284252f63dc1dcc563da038fce76)), I'm still seeing it pull the python slim image every time. Perhaps I've implemented the `build-push-action` incorrectly, but from my understanding it doesn't cache the final image (only the build layers), which is potentially why it's pulling python slim every time? 

In this iteration, I re-worked the RL to push to GHCR as `rl-service:pr-$PR_NUMBER` so that we don't run into conflicts with `rl-service:latest` when two builds happen simultaneously as you pointed out on the other PR. Thoughts?

I'm also happy to pair on this if you have time and some additional thoughts on the `build-push-action` tact.

m-goggins:
gha caching does not persist across jobs. I forgot to mention that I tried both (caching and building in one job and separately) but neither seemed to do the trick.

 I'll proceed with the `rl-service:pr-$PR_NUMBER` solution and write a follow up ticket to handle the clean up.

ericbuckley:
> gha caching does not persist across jobs. I forgot to mention that I tried both (caching and building in one job and separately) but neither seemed to do the trick.
> 
> I'll proceed with the `rl-service:pr-$PR_NUMBER` solution and write a follow up ticket to handle the clean up.

@m-goggins thanks for all your experimentation here, testing is particularly tricky as we haven't run into a dockerhub authentication issue yet!  Another option to consider, the work you did on the database services gets us more than halfway there (according to my count, we use 6 service images in every commit check and 4 rl-service images).  I'm wondering, if maybe we just roll out the database service caching work you did and call it at that.  Wait to see if anything breaks post Apr 1st and go from there.

ericbuckley (approved):
This looks good @m-goggins.  One last question on the schedule job, did you check to see that it runs as expected? I think as is, it will only run on April 1st.  Wondering if we want to change up the "on" schedule in a commit, just to verify that it works, then flip it back.

m-goggins:
I tried testing on this branch, but it looks like scheduled jobs only run on the [default branch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) of a repo so I can't test until it's merged into main. 

I set it to run at 12:01am on the 27th so assuming you're okay with merging, I can check to see if it worked tomorrow and re-adjust to the 1st if everything goes smoothly.

ericbuckley:
> I tried testing on this branch, but it looks like scheduled jobs only run on the [default branch](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) of a repo so I can't test until it's merged into main.
> 
> I set it to run at 12:01am on the 27th so assuming you're okay with merging, I can check to see if it worked tomorrow and re-adjust to the 1st if everything goes smoothly.

Oh, I meant instead of changing the schedule, change the trigger.  The below would have the workflow run on a PR that is trying to merge into main.

```
on:
  push:
    branches: [main]
```

m-goggins:
Oh, yep! In that case, it has run successfully, here: https://github.com/CDCgov/RecordLinker/actions/runs/14065686730

ericbuckley (approved):

PR #269, opened 2025-03-31
normalize telecom model_validator

### LINKED ISSUE #248: validate and normalize telecom data
## Summary
Normalize and validate all telecom data upon receipt, instead of when emitting the values for evaluation.

## Acceptance Criteria
- [x] Remove Telecom.phone and Telecom.email methods
- [x] Add a Telecom model validator to validate and normalize phone numbers
- [x] Add a Telecom model validator to validate email addresses (no need to normalize)
- [x] Update feature_iter to emit the appropriate email/phone values
- [x] Update site/reference.md to give guidance on how phone is normalized

## Details / Tasks
For phone number, please consider using [phonenumbers](https://pypi.org/project/phonenumbers/) or another lib to parse the incoming number and produce a [E164](https://en.wikipedia.org/wiki/E.164) string.

Validate email addresses with an email regex.

Both email addresses and phone numbers should use the `normalize_text` method in the feature_iter.

m-goggins:
## Description
This PR adds a model validator to the `Telecom` class that strips leading/trailing whitespace from emails and normalizes phone as an E164 string, i.e., including country codes but excluding extensions. When using `feature_iter`, only the national number (no country code) is used for comparison to avoid issues with Jaro-Winkler over indexing on similarities in country codes. For emails, we do not apply text normalization in `feature_iter` so that we can preserve the special characters for comparison. 

Storing:
- email: email with leading/trailing whitespace removed
- phone number: normalized to E164 where possible, otherwise store raw number

Comparing:
-email: same as storage
phone number: national number (no country code) and normalized_text() is applied to remove additional special characters. I primarily kept normalized_text in case we were not able to normalize to E164 during ingestion and we want to remove special characters.

## Related Issues
Fixes #248 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/tests/unit/schemas/test_pii.py
+++ b/tests/unit/schemas/test_pii.py
@@ -378,23 +378,6 @@ class TestPIIRecord:
             "555",
         ]
 
-    def test_feature_iter_telecom_phone(self):
-        record = pii.PIIRecord(
-            telecom=[
-                pii.Telecom(value="+1 555-123-4567", system="phone"),
-                pii.Telecom(value="+15551234567", system="phone"),
-                pii.Telecom(value="555-987-6543 ext 123", system="phone"),
-                pii.Telecom(value="555", system="phone"),
-            ]
-        )
-
-        assert list(record.feature_iter(pii.Feature(attribute=pii.FeatureAttribute.TELECOM))) == [
-            "5551234567",
-            "5551234567",
-            "5559876543",
-            "555",
-        ]
-
     def test_blocking_keys_invalid(self):
         rec = pii.PIIRecord()
         with pytest.raises(ValueError):


src/recordlinker/schemas/pii.py
```diff
@@ -234,36 +234,31 @@ class Telecom(StrippedBaseModel):
...
-            name: name
-            for name, method in cls.__dict__.items()
-            if callable(method) and not name.startswith("_")
-        }
+    @pydantic.model_validator(mode="before")
+    def validate_and_normalize_telecom(cls, values):
+        """
+        Validate and normalize the telecom record.
+        """
+        # If telecom.system = "email", set telecom.value to lowercase
```
ericbuckley, line 242: Based on the comment, I think we're missing a call to `lower()`

src/recordlinker/schemas/pii.py
```diff
@@ -441,27 +438,31 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
                     yield str(race)
         elif attribute == FeatureAttribute.TELECOM:
             for telecom in self.telecom:
+                value = telecom.value.strip()
```
ericbuckley, line 441: I don't think we need to call `strip()`, all whitespace should have been removed by the model_validator

src/recordlinker/schemas/pii.py
```diff
@@ -441,27 +438,31 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
...
-                handlers = Telecom.get_system_handlers()
-
-                if telecom.system in handlers:
-                    value = getattr(telecom, handlers[telecom.system])()
-                    if value:
-                        yield value
-
+                elif telecom.system == "email":
+                    yield value
+                elif telecom.system == "phone":
```
ericbuckley, line 447: I think we just need 1 if/else block and no elif blocks here.  system == None or system == "email" is exactly the same as the else block.

m-goggins: Ah you're right; forgot to clean it up after making other changes. Good catch!

src/recordlinker/schemas/pii.py
m-goggins: Oops, you're right

src/recordlinker/schemas/pii.py
```diff
@@ -234,36 +234,31 @@ class Telecom(StrippedBaseModel):
...
+        Validate and normalize the telecom record.
+        """
+        # If telecom.system = "email", set telecom.value to lowercase
+        if values.get("system") == "email":
+            values["value"] = values["value"].strip().lower()
+        # If telecom.system = "phone", normalize the number
+        elif values.get("system") == "phone":
+            try:
+                # Attempt to parse with country code
+                if values["value"].startswith("+"):
```
ericbuckley, line 249: This is going to raise a server error if the user doesn't send the data in the correct format.  For example start up the local_server.sh and run this.

```sh
curl -X POST -H "Content-Type: application/json" -d '{"record": {"telecom": [{"system": "phone","val": " 555-123-4567"}]}}' http://localhost:8000/link
```

src/recordlinker/schemas/pii.py
```diff
@@ -441,27 +438,25 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
...
-
-                if telecom.system in handlers:
-                    value = getattr(telecom, handlers[telecom.system])()
-                    if value:
-                        yield value
-
+                value = telecom.value
+                if telecom.system == "phone":
+                    # Use national number for comparison
+                    phone = normalize_text(str(phonenumbers.parse(telecom.value).national_number))
```
ericbuckley, line 444:
```suggestion
                    phone = normalize_text(str(phonenumbers.parse(value).national_number))
```
Alternatively, the `value` local doesn't add much.  you could just change line 448 to `yield telecom.value`

ericbuckley (changes requested):

src/recordlinker/schemas/pii.py
m-goggins: Ah, I switched the mode to "after", which should take care of this issue the model validator will now run after the whole model has been validated.

src/recordlinker/schemas/pii.py
```diff
@@ -234,36 +234,32 @@ class Telecom(StrippedBaseModel):
...
+        # If telecom.system = "email", set telecom.value to lowercase
+        #
+        if self.system == "email":
+            self.value = self.value.strip().lower()
+        # If telecom.system = "phone", normalize the number
+        elif self.system == "phone":
+            try:
+                # Attempt to parse with country code
+                if self.value.startswith("+"):
+                    parsed_number = phonenumbers.parse(self.value)
```
ericbuckley, line 251: I don't think we're doing anything with the `parsed_number` variable.  Shouldn't we be assigning self.value to it?

src/recordlinker/schemas/pii.py
```diff
@@ -441,27 +439,24 @@ def feature_iter(self, feature: Feature) -> typing.Iterator[str]:
...
-                handlers = Telecom.get_system_handlers()
-
-                if telecom.system in handlers:
-                    value = getattr(telecom, handlers[telecom.system])()
-                    if value:
-                        yield value
-
+                if telecom.system == "phone":
+                    # Use national number for comparison
+                    phone = normalize_text(str(phonenumbers.parse(telecom.value).national_number))
```
ericbuckley, line 444: I understand why we are calling `normalize_text` here, but why are we reparsing the value with phonenumbers?  Shouldn't we only have to do that once during in the `validate_and_normalize_telecom` method?

ericbuckley: Ah thanks for that explanation, that's helpful!

ericbuckley (approved):
🚀

tests/unit/schemas/test_pii.py
```diff
@@ -583,3 +601,20 @@ def test_parse_state(self):
         assert address.state == "Armed Forces"
         address = pii.Address(state="Conneticut")
         assert address.state == "Conneticut"
+
+
+@pytest.mark.parametrize(
```
ericbuckley, line 606: neat!

ericbuckley (approved):

PR #270, opened 2025-03-31
remove Patient.record circular dependency

### LINKED ISSUE #176: Fix Patient.record dependency
## Summary
There is a circular dependency between models/mpi and schemas/pii.  The logic in the mpi Patient.record methods should be removed so we don't have to tip toe around that dependency.

ericbuckley:
## Description
Removing the circular dependency between `models/mpi.py` and `schemas/pii.py`.  `pii.py` now depends on `mpi.py`, but not vice-versa.

## Related Issues
closes #176 

## Additional Notes
The circular dependency didn't cause any issues, as the mpi.py module imported all of its pii.py dependencies in functions rather than at the top of the module, but this is a hack.  Having two modules depend on each other opens potentials for errors during refactors, or could lead to more spaghetti code if attempt to use PIIRecord in mpi.py for other uses.

- pii.PIIRecord created 2 new methods for converting a Patient into a PIIRecord and back.  PIIRecord.from_patient and PIIRecord.to_data.
- To reduce the number of times in which we convert a Patient object to a PIIRecord, the matching function signatures have been updated to accept two PIIRecord for comparison, rather than a PIIRecord and Patient.

## Checklist

## Checklist for Reviewers


bamader (approved):
Good catch! As an added bonus, I like the `from_patient` function in schemas now more than `model_construct` (pure semantics, I know, but `from_X` is a common functional construct so I think it's more readable, just gravy).

PR #272, opened 2025-03-31
Add AlgorithmPass label and description fields

### LINKED ISSUE #246: Algorithm Pass label and description
## Summary
Algorithm Pass should have a label and description field to assist users identifying which pass they are viewing/editing.

## Acceptance Criteria
- [ ] New fields on the AlgorithmPass model for label & description
- [ ] New fields on the AlgorithmPass schema for label & description
- [ ] Updates to initial_algorithms.json with new pass labels

## Details / Tasks
Use the same label, description fields on the Algorithm model for AlgorithmPass.  The label should be required, but only has to be unique **per Algorithm** (not across the whole table).  The label can default to an auto-generated value based on the evaluators selected.  Description can remain as an optional text field.

## Background / Context
We observed the NBS mod designing for this during a meeting.

## Notes / Comments
We'll need to add a note in our release documentation to indicate this is another breaking change.

ericbuckley:
## Description
Add label and description fields to Algorithm Pass to assist users in identifying which pass they are viewing/editing.

## Related Issues
closes #246 

## Additional Notes
Both of these fields are optional, in the event that a label isn't specified we'll default to a value that is based on the evaluators used.  This wasn't specified in the spec, but I thought it be nice to auto generate some labels for users.  I believe Henry was doing something similar in the mockups he put together for NBS.

## Checklist

## Checklist for Reviewers


src/recordlinker/schemas/algorithm.py
```diff
@@ -54,6 +59,16 @@ class AlgorithmPass(pydantic.BaseModel):
     rule: matchers.RuleFunc
     kwargs: dict[str, typing.Any] = {}
 
+    @pydantic.model_validator(mode="after")
+    def default_label(self) -> "AlgorithmPass":
+        """
+        Create a default label for the algorithm based on the keys used in the evaluation step.
+        """
+        if not self.label:
+            keys = [str(e.feature) for e in self.evaluators] + ["match"]
```
ericbuckley, line 68: @m-goggins @bamader interested in your thoughts on this function.  I thought a default label might be helpful for users when creating passes. If you agree, is this the right default, should it be based on evaluation features?

m-goggins: I like the idea of a default, though I would add blocking values + 'block' in addition to evaluators +'match'

ericbuckley: I see how that adds value, but I think we should contrast that with the length of the label.  I feel if the label is too long its hard for a user to quickly grok what the pass is.  For example, using our dibbs-default pass 1.  Using just evaluators we have a default label of `"FIRST_NAME-LAST_NAME-match"` (26 chars).  Using both blocking and evaluators we have a default label of `"BIRTHDATE_IDENTIFIER_SEX_FIRST_NAME_LAST_NAME-match"` (51 chars).  To me, the second is a little harder on the eyes, but maybe thats not the same for others?

m-goggins: My preference is for the latter because it's more complete, but would make it clearer which values are for blocking and which are for matching: `block_BIRTHDATE_IDENTIFIER_SEX_match_FIRST_NAME_LAST_NAME`. I also think we can pick one and get feedback from users. I don't want this to be a blocker for this work so I'll approve.

m-goggins (dismissed):

src/recordlinker/schemas/algorithm.py
ericbuckley: sounds good, a question for Matt maybe?

bamader: I think the idea of a default label is a great one, but I like the second option as well (I tend to prefer long descriptions and names in general, in both comments and variables, because I think it helps stop confusion before it starts and can even create informal standards or practices around labeling within code--for example, when I was working on the missingness stuff, Marcelle recognized we'd created the norm of using `_proportion` to refer to 0 to 1 floats that specifically describe an amount of something else, so if we create such a parameter in the future we now have an established naming convention for it; not super relevant to this example, I just bring it up to elaborate on my general preference). My suspicion is that users will provide their own labels most of the time and so this default will get overridden, but as a compromise idea, what if we inverted the casing of the `block_` and `match_` tags and the feature field names to allow the eyes to visually pick out the section of interest (via caps) and skim over details they can pass on unless they want to read (specific fields, via lower case)?

So in the example Marcelle gave, we would have `BLOCK_birthdate_identifier_sex_MATCH_first_name_last_name`. It's softer on the eyes but also still complete, and since we don't use any part of the label to index into a dict the casing doesn't have to match the feature precisely. Thoughts?

bamader (approved):
Left some thoughts on the default name, but don't want that to obstruct anything so giving my +1

src/recordlinker/schemas/algorithm.py
ericbuckley: I'll give it a shot

ericbuckley: @m-goggins @bamader changed.   Let me know what you think https://github.com/CDCgov/RecordLinker/pull/272/commits/e6d9c84badbd806e76535252ec1b8de57997dc8a

src/recordlinker/assets/initial_algorithms.json
```diff
@@ -54,6 +55,7 @@
                 }
             },
             {
+                "label": "address-birthdate-match",
```
m-goggins, line 58: ditto BLOCK_x_x_MATCH_y_y

m-goggins (dismissed):
Looks good, just want to make sure the initial algorithms match the defaults! Thanks for making the changes; I like Brandon's suggestion to make it easier to read while still capturing the information for users.

src/recordlinker/assets/initial_algorithms.json
ericbuckley: done. https://github.com/CDCgov/RecordLinker/pull/272/commits/1b35621a3b905e362f981ebc2da18e18e61ed099

m-goggins (approved):
🚀

PR #282, opened 2025-04-08
bugfix for null values

### LINKED ISSUE #281: null linkage features create server errors
## Summary
When sending data to the `/link` and `/match` endpoints, "null" values in specific feature attributes cause server errors.

## Impact
Invalid user data should either be ignored, or raise a 422 and specify to the user what was incorrectly specified.  Letting a server error propagate through doesn't give the user enough context to know if the problem is on their end or with the system.

## Steps to reproduce
The following values in the "record" input (used in `/link`, `/match` and `/seed`) all cause server errors.

- `"race" = null`
- `"address": [{"line": null}]`
- `"identifiers": {"type": null, "value": "X"}`

## Expected behavior
In the first two cases (race and address.line) we should just ignore the null and accept that payload as if the values weren't sent.  In the last example, we should raise a 422 and explicitly state the identifier type is a required value.

ericbuckley:
## Description
Fix to address server errors when a null value is send for `race`, `identifier.type` or `address.line`.

## Related Issues
closes #281 

## Additional Notes
Also fixing a broken test case from a bad merge.  https://github.com/CDCgov/RecordLinker/actions/runs/14340832961

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/schemas/identifier.py
+++ b/src/recordlinker/schemas/identifier.py
@@ -1,5 +1,4 @@
 import enum
-import re
 import typing
 
 import pydantic
@@ -9,6 +8,7 @@ class IdentifierType(enum.Enum):
     """
     Enum for various identifier types.
     """
+
     AC = "AC"
     ACSN = "ACSN"
     AIN = "AIN"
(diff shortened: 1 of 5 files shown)


m-goggins (commented):
Looks good for these fields! Are you concerned about null values in other fields?

bamader (dismissed):
Seems legit 👍 . Similar to Marcelle, were these the only fields that Michael identified as having a problem?

ericbuckley:
@m-goggins @bamader good questions.  Michael only reported the issue on `race = null`, I discovered the other two when reading the code in pii.py.  From what I can tell, all the fields that are validated by pydantic using types (eg name.family) seem to handle null values just fine.  I focused my testing on all the fields with custom field_validator functions, specifically the ones executed before validation, as there is little guarantee as to what the value is at that time.

If you think there could be issues with other fields, let me know, and I'll add a quick test case for that.

m-goggins (approved):
Looks good!

PR #291, opened 2025-04-14
process skip values

### LINKED ISSUE #233: Add cleaning step to process skip_values
## Summary
Implement a new cleaning step to copy PIIRecords with `skip_values` removed as per [RFC-003](https://github.com/CDCgov/RecordLinker/pull/218).

## Acceptance Criteria
- [ ] A new cleaning method for PIIRecords that can copy an existing object, parse some skip rules and return a new one.
- [ ] Apply cleaning method to both incoming payloads and selected patients from blocking results
- [ ] Test cases

## Details / Tasks
Consider implementing this new method somewhere in the "recordlinker/linking" module as to avoid a circular dependency between schema definitions.  Also take care to invoke the cleaning step on both the incoming payload and found patient records in the MPI as early as possible, so we don't call the cleaning operation more than what's needed.

The Patient class in models/mpi.py as some [tricky circular dependencies defined on the record property](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/models/mpi.py#L79-L106).  This maybe be a good opportunity to refactor those circular dependencies, if so, feel free to tackle that work here.

## Dependencies
#232.  Largely this work isn't dependant, however the work from #232 will likely create some new schema object from which a list of skip value definitions are defined.  This work is necessary for the cleaning method, as it will need to know how to process that schema.

## Related Issues/PRs
#215

ericbuckley:
## Description
Add functionality to clean data of all skip values before blocking or evaluation.

## Related Issues
closes #233 

## Additional Notes
- There is some duplication between `PIIRecord.feature_iter` and the new clean method, but it's not obvious to me that abstracting some of that functionality is a gain over the readability cost to understanding how both methods work.  I'm very much open to other options, if others have thoughts on how we could combine logic from these two methods.
- The new `clean` method was purposefully put into the `linking` module to remove the potential for a circular import between `schemas.algorithm` and `schemas.pii`
- Using `fnmatch` over a `regex` has some real downsides when matching on "John Doe".  I think our default algorithm config is going to end up have 6 values (eg `["John Doe", "John * Doe", "Jon Doe", "Jon * Doe", "Jane Doe", "Jane * Doe"]).  Open to hear if people think this is a good reason to switch to a regular expression, or maybe we support both somehow?

## Checklist

## Checklist for Reviewers


ericbuckley:
@m-goggins @bamader we never discussed making changes to `dibbs-default` based on this work.  Should we add a default "skip_values" list in this PR, maybe in another or should we just leave it blank and let users decide?

src/recordlinker/linking/link.py
```diff
@@ -126,6 +130,8 @@ def compare(
       like which fields to evaluate and how to total log-odds points.
     :param log_odds_weights: A dictionary mapping Field names to float values,
       which are the precomputed log-odds weights associated with that field.
+    :param skip_values: A list of SkipValue objects that contain information
+      about fields that should be skipped during comparison.
```
m-goggins, line 134:
```suggestion
      about field values that should be skipped during comparison.
```

src/recordlinker/linking/link.py
```diff
@@ -134,8 +140,11 @@ def compare(
     evals: list[models.BoundEvaluator] = algorithm_pass.bound_evaluators()
     # keyword arguments to pass to comparison functions
     kwargs: dict[typing.Any, typing.Any] = algorithm_pass.kwargs
-    # convert the Patient model into a PIIRecord for comparison
-    mpi_record: schemas.PIIRecord = schemas.PIIRecord.from_patient(patient)
+    # convert the Patient model into a cleaned PIIRecord for comparison
+    # FIXME: can we clean higher up on the stack?
+    mpi_record: schemas.PIIRecord = clean.clean(
+        schemas.PIIRecord.from_patient(patient), skip_values
+    )
```
m-goggins, line 147: I think we could move the cleaning to `link_record_against_mpi` (line 269) where we are iterating through `pat in pats`. We then wouldn't have to pass skip values to `compare` and instead just pass `cleaned_record` and `cleaned_mpi_record` which would now also be a PIIRecord instead of `models.Patients`

src/recordlinker/linking/link.py
```diff
@@ -235,13 +249,13 @@ def link_record_against_mpi(
...
+            # block on the cleaned_record and the algorithm's blocking criteria, then
             # iterate over the patients, grouping them by person
             with TRACER.start_as_current_span("link.block"):
                 # get all candidate Patient records identified in blocking
                 # and the remaining Patient records in their Person clusters
                 pats = mpi_service.BlockData.get(
-                    session, record, algorithm_pass, max_missing_allowed_proportion
+                    session, cleaned_record, algorithm_pass, max_missing_allowed_proportion
                 )
                 for pat in pats:
```
m-goggins, line 260: You could also move the removal of skip values here instead of line 269/270 as I suggested before.

m-goggins (commented):
Thanks for implementing, this is looking pretty good to me!

> There is some duplication between PIIRecord.feature_iter and the new clean method, but it's not obvious to me that abstracting some of that functionality is a gain over the readability cost to understanding how both methods work. I'm very much open to other options, if others have thoughts on how we could combine logic from these two methods.
I think this is okay for now; it's definitely readable as is, and we can always re-evaluate in the future if the duplicate code becomes more burdensome to maintain.

> The new clean method was purposefully put into the linking module to remove the potential for a circular import between schemas.algorithm and schemas.pii
I'm fine to leave in the linking module but I the names "clean" and "matches" are confusing in the context of the rest of record linkage. I would suggest we make it more specific to skip values like `remove_skip_values` instead of `clean` and `matches_skip_values` instead of `matches`. 

> Using fnmatch over a regex has some real downsides when matching on "John Doe". I think our default algorithm config is going to end up have 6 values (eg `["John Doe", "John * Doe", "Jon Doe", "Jon * Doe", "Jane Doe", "Jane * Doe"]). Open to hear if people think this is a good reason to switch to a regular expression, or maybe we support both somehow?

I think we could make it easier on users and if they provide "John Doe" for NAME, we generate the wild cards for them with a helper function so that they don't have to be familiar with the wild cards, only the specific skip values. Thoughts?

src/recordlinker/linking/link.py
ericbuckley: I want to do something like that, but the problem right now is we're still using the MPI patient model in this compare method (see line 151).  However, we are using it just for logging, so if we remove the extra logging detail we could rewrite the compare method to accept two PIIRecords rather than a PIIRecord (incoming) and Patient (existing) object.  What do you think, remove the logging to speed things up a bit?

m-goggins: I want to make sure I understand the purpose of logging `Patient.reference_id`. We're doing that so that we can track all of the Patients an incoming record was compared to, right, not just the final one it was linked to?

ericbuckley: Yeah, it's just debugging info to help with a potential investigation if we ever wanted to dig into the details of how/why results were being returned.

m-goggins: In that case, I think it is okay to remove so that both `record` and `patient` are PIIRecords.

ericbuckley: cool, I'll give this a try

ericbuckley: So I just made the change, https://github.com/CDCgov/RecordLinker/pull/291/commits/af509f0d284445c64209eaf54db87b4084e116ad, but I'm rethinking its value.

With multiple passes, its possible that there will be some overlap in the MPI records returned.  In fact, we highlighted this in some of the examples we put together.  However, the way its implemented now, we have to "re-clean" the MPI Patient data everytime its pulled back from the database (so every pass).  In other words, I think we end up calling the clean method the same number of times, just in a different place.

I think there is an upside to the new approach though, as now all cleaning is done in one method, `link` and it's not split between the two.

m-goggins: Gotcha. I think the consolidation to one method and passing in the same data type (PIIRecord) seem like improvements to me.

m-goggins:
Thoughts on my other feedback?

> The new clean method was purposefully put into the linking module to remove the potential for a circular import between schemas.algorithm and schemas.pii

I'm fine to leave in the linking module but I the names "clean" and "matches" are confusing in the context of the rest of record linkage. I would suggest we make it more specific to skip values like remove_skip_values instead of clean and matches_skip_values instead of matches.

> Using fnmatch over a regex has some real downsides when matching on "John Doe". I think our default algorithm config is going to end up have 6 values (eg `["John Doe", "John * Doe", "Jon Doe", "Jon * Doe", "Jane Doe", "Jane * Doe"]). Open to hear if people think this is a good reason to switch to a regular expression, or maybe we support both somehow?

I think we could make it easier on users and if they provide "John Doe" for NAME, we generate the wild cards for them with a helper function so that they don't have to be familiar with the wild cards, only the specific skip values. This might also be worth another ticket.

src/recordlinker/linking/clean.py
```diff
@@ -0,0 +1,122 @@
+"""
+recordlinker.linking.clean
```
ericbuckley, line 2: @m-goggins starting a thread here to discuss your comment

> I'm fine to leave in the linking module but I the names "clean" and "matches" are confusing in the context of the rest of record linkage. I would suggest we make it more specific to skip values like remove_skip_values instead of clean and matches_skip_values instead of matches.

- On the module name, are you good with `recordlinker.linking.clean` or do you think we should adjust?
- I'm fine with renaming the `clean` method to `remove_skip_values`, I'll do that today.
- I went back and forth on making `matches` private, as really this is just an internal method to dry up the code in the clean function and I don't think external modules should use.  I'm wondering if I should make that change as well.  So maybe rename to `_matches` or `_match_skip_values`?

src/recordlinker/linking/clean.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Return whether the value matches any of the values in the list
+    using case-insensitive fnmatch matching.
+
+    :param value: the value to match
+    :param values: the list of values to match
+    :return: True if the value matches any of the values in the list
+    """
+    val = value.lower()
+    return any(fnmatch.fnmatch(val, v.lower()) for v in values)
```
ericbuckley, line 25: @m-goggins starting a thread here to discuss this comment

> I think we could make it easier on users and if they provide "John Doe" for NAME, we generate the wild cards for them with a helper function so that they don't have to be familiar with the wild cards, only the specific skip values. This might also be worth another ticket.

I 100% agree we should make this easier for users to specify.  Having to write 6 values for our John/Jane Doe examples is a bit much for users.  If possible, I'd like to stay away from creating magic values that evaluate to a different pattern than others.  We could put in documentation to explain this, but I'm worried that some users won't see that documentation and get confused with strings like "John Doe" match patterns differently than "123 Main St".

I wonder if we could extend the fnmatch syntax with an extra wildcard like `+` or maybe do some special processing with all space characters?  Alternatively, we could opt for an advanced mode where users are allowed to provide a regex (eg `"Jo[h|n]n\s\*?\sDoe"`)?

src/recordlinker/linking/clean.py
m-goggins: I think we should rename the module too, to something that gets at skip values. For the internal method, I like `_match_skip_values`

src/recordlinker/linking/clean.py
m-goggins: I'm trying to think of examples of when we'd want the skip values to be evaluated differently based on field but having a hard time. If there's a skip value like "123 Fake Street," we'd also want to remove "123 Fake Street, Apt 2" right? Did NBS ever provide a list of skip values? I think it would be helpful to see some examples.

in the absence of examples, I'd caution against regex as an advanced option and opt for creating magic values that evaluate to a different pattern than others (plus documentation). Though it's less than ideal from a documentation perspective, with the option I proposed, we wouldn't run the risk of a regex option that accidentally treats all characters as skip values.

src/recordlinker/linking/clean.py
bamader: Slightly late here, but +1 to renaming the module as well, both for clarity of purpose (handling skip_values) and clarity for the users.

I'm good with `_match_skip_values` on the internal name!

src/recordlinker/linking/clean.py
bamader: I would personally opt pretty strongly against an advanced mode and regex as a whole. Our users are going to be smart people, especially the admins who toggle the algorithm, but they're statisticians, not developers, and that's a distinction we've learned in the past can be very significant. Regex and wildcards are not familiar concepts to them, and even with documentation it will create a larger than needed mental load on them; they _will_ think to themselves something to the effect of "Why do I have to type Jo[h|n]n\s\*?\sDoe to represent an easy name everyone in my field knows?" 

I think the user friendliness of magic values with different evaluation patterns is worth the less than ideal software design principle it carries. I think it will streamline things for the user in a way they're more likely to engage with and will increase their comfort level working with algorithm configuration (because the more of the toggles they understand, the more they'll trust it).

Marcelle, to your question about the 123 Fake Street, I would definitely think we want to remove modifications or additional information in a field once a skip value is detected. E.g., 123 Fake Street Apartment 2 is a skip value, just as 123 Fake Street NW is, just as 123 Fake Street on its own is. I don't know _why_ there would be appended data, but yeah I feel like once the skip value root is there the field is done for anyway.

docs/developer_guide.md
```diff
@@ -54,6 +54,10 @@ following steps should be taken:
 ### Modify the `PIIRecord.feature_iter` Method
 - Update the [PIIRecord.feature_iter](https://github.com/CDCgov/RecordLinker/blob/a672d2b6409cbd1a08f729d94fba5692f57f6fc6/src/recordlinker/schemas/pii.py#L246) method to return the value of the new feature when it's used for comparison.
 
+### Modify the `linking.clean.clean` Function
+- Update the `recordlinker.linking.clean.clean` function to process and remove any data
+that matches the new feature.
+
```
bamader, line 60: (just flagging this will need changed when we update the new name)

src/recordlinker/linking/clean.py
```diff
@@ -0,0 +1,122 @@
+"""
+recordlinker.linking.clean
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This module is used to clean the data before running the linkage algorithm
```
bamader, line 5: Nit: when we change the module name, I would also alter the copy of this comment to reflect that it's concerned with removing skip-values, until/unless we put more cleaning functions in

bamader (commented):
Code looks mostly good--just the couple of things around module naming and modes, I think.

src/recordlinker/linking/clean.py
ericbuckley: Ok I hear you both on the regex.  The magic values still scare me a bit from a user perspective, as when they add a string we didn't explicitly write a special condition for, the behavior changes for them.  I realize most will type in "John Doe", but as soon as one of them writes "Johnathan Doe" are magic string code fails to work.

Taking a second look at the original 6 values I had to add (`["John Doe", "John * Doe", "Jon Doe", "Jon * Doe", "Jane Doe", "Jane * Doe"]), I think the annoying part for users is having to write "John Doe" and "John * Doe", to catch examples when there is a middle name and when there isn't one.

What if we wrote a special rule, such that if we find a string with an `*` wildcard surrounded by whitespace.  We add a second value where the `*` is removed and the whitespace is collapsed.  For example:
- "John * Doe" -> ["John Doe", "John * Doe"]
- "Test *    Value" -> ["Test Value", "Test *     Value"]

m-goggins: So users would only have to enter "John * Doe"? I like limiting what users have to input so that it's easier, although I still think adding "John Doe" is more intuitive. 

I think "John Doe" and "Johnathan Doe" should behave differently so I'm not sure I'm following you on that example.

ericbuckley: @m-goggins @bamader as we discussed earlier, I have removed the [fnmatch](https://docs.python.org/3/library/fnmatch.html) matching syntax from skip values.  Let me know what you think.

https://github.com/CDCgov/RecordLinker/pull/291/commits/8f3a78193601b4f015c6bcfc3282a8985246e13a

m-goggins (approved):
Looks good to me! I'm glad we have an approachable way to handle skip values to start and can always make changes based on user feedback.

PR #320, opened 2025-04-18
configurable API root path

### LINKED ISSUE #287: configurable API root path
## Summary
Add a setting that allows users to configure the default API path.

## Acceptance Criteria
- [ ] Add a new setting to the config, API_ROOT_PATH
- [ ] Add validations for the API_ROOT_PATH
- [ ] Update the main app to use the [default root_path](https://fastapi.tiangolo.com/reference/fastapi/?h=fastapi#fastapi.FastAPI--example)

## Details / Tasks
This is a request coming from NBS as they would like to have the API paths in their K8 cluster more closely align with patterns of their other APIs.  Specifically, NBS is hoping to set the above API_ROOT_PATH value to "/recordlinker/api".  When validating, we need to verify that the path is NOT the root path "/".  Validating should also clean the incoming value so that the leading and trailing slashes are normalized.  We don't want to host our API from "localhost:///recordlinker/api/health-check", nor do we want urls like "localhost://recordlinker/api//health-check".  Basically make sure we don't have a double "/".

ericbuckley:
## Description
Adding a configuration setting to change the root path of the API

## Related Issues
closes #287 

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -64,12 +64,15 @@ jobs:
           docker run -d --name rl-service \
             --network="host" \
             -e DB_URI=$DB_URI \
+            -e API_ROOT_PATH="/api" \
             rl-service-image
+
+          BASE_URL="http://localhost:8080/api"
 
           # Wait for the RL Service to be healthy
           TRIES=5
           COUNT=0
-          until curl -s http://localhost:8080/ | grep "OK"; do
+          until curl -s $BASE_URL | grep "OK"; do
             echo "Waiting for RL Service to become healthy... Attempt $((COUNT+1)) of $TRIES"
             sleep 5
             COUNT=$((COUNT+1))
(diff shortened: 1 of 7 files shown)


.github/workflows/check_smoke_tests.yml
```diff
@@ -64,12 +64,15 @@ jobs:
           docker run -d --name rl-service \
             --network="host" \
             -e DB_URI=$DB_URI \
+            -e API_ROOT_PATH="/api" \
             rl-service-image
 
+          BASE_URL="http://localhost:8080/api"
+
```
bamader, line 71: More a curiosity than anything, but in this case do the BASE_URL and API_ROOT need to have the same extension? I.e. if the BASE_URL were `localhost:8080` but then the API_ROOT were `/api` would that cause problems? Wondering mostly because as written this seems to move exposure to the whole app behind the API path, which may or may not be the intended functionality. My intuition is that anything prefixed with `/api` is something that an API-specific consumer would access--such as some pipeline system reaching out for a request--whereas the root of `localhost:8080/` (no `/api` included) would be used by e.g. us or a user messing around on their local with the app open. Is that right?

bamader (approved):
Everything looks good! Just have a question about what parts of the app we're specifically trying to situate behind the `/api` tag, but it's non-blocking so giving the +1.

.github/workflows/check_smoke_tests.yml
ericbuckley: There is a dependency between the two, and maybe its better to create two variables to show that.
```shell
API_ROOT="/api"
BASE_URL="http://localhost:8080${API_ROOT}"
```

The intent is that all API calls and at least access to the openapi schema (`/api/openapi.json`), will be behind the prefix.  Micheal didn't explicitly state whether the docs (swagger and redoc) should be behind this root path or not, but to me, it made sense to do that as well (this is certainly open for discussion if you disagree).

In `main`, the root URL isn't be used for anything and is currently returning a 404.  I'm wondering if we should add a redirect to the docs?  Thoughts?

bamader: Yeah no strong feelings on whether the docs should be behind the root, I could see good justification either way. good find on the root URL returning a 404, though--I think redirecting to the docs removes what might seem like a "failure point" to a potential user and keeps them firmly directed towards our instruction and education.

PR #322, opened 2025-04-21
Add suffix normalization

### LINKED ISSUE #297: Add suffix normalization
## Summary
We need to normalize suffixes upon ingestion so that they may be prepended.

## Acceptance Criteria
- [ ] A Pydantic `field_validator` is added to the Name feature that normalizes suffixes
- [ ] Non-roman numerals should ideally be shortened, e.g., Senior -> Sr 
- [ ] Suffixes that cannot be normalized should be returned as is
- [ ] Documentation is updated to reflect that normalization occurs

## Details / Tasks
- I'm not sure if we should normalize roman numerals at this stage; it may be beneficial to preserve the numerals for the sake of readability to an epi looking at records (Richard 3 looks odd compared to Richard III). We could only normalize roman numerals in blocking fields and during evaluation of FULL_FIRST_NAME to abstract this oddness away from users.
- It may be helpful to have a list of common suffixes, but we could start with Senior, Sr, Junior, Jr

bamader:
## Description
This changeset adds a pydantic validator that parses and normalizes suffixes according to a predefined mapping. I've left the mapping in an assets file in case we want to add roman numeral standardization at some later date.

## Related Issues
Fixes #297 

## Additional Notes
I did some research on common suffixes in names in the US, and most results indicated it's pretty limited, just Junior, Senior, II, and III. It's easy to support roman numerals, as well as any other spelling or punctuated variants thereof, but given this I think leaving the suffixes we intend to support in an assets file (like with state abbreviations and street conventions) follows our established practice.

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -19,6 +19,9 @@ from recordlinker.utils.normalize import normalize_text
 _STATE_NAME_TO_CODE = utils.read_json("assets/states.json")
 _STATE_CODE_TO_NAME = {v: k for k, v in _STATE_NAME_TO_CODE.items()}
 
+# Load suffix mappings for Name normalization
+_SUFFIX_VARIANTS_TO_STANDARD_SUFFIXES = utils.read_json("assets/suffixes.json")
+
 
 class FeatureAttribute(enum.Enum):
     """
@@ -60,6 +63,7 @@ class StrippedBaseModel(pydantic.BaseModel):
         if isinstance(v, str):
             return v.strip()
         return v
+    
 class Feature(StrippedBaseModel):
     """
     The schema for a feature.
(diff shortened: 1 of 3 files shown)


src/recordlinker/schemas/pii.py
```diff
@@ -182,6 +186,23 @@ class Name(StrippedBaseModel):
...
+        """
+        Parse and normalize the suffix field into a standard representation.
+        """
+        normalized: list[str] = []
+        if value:
+            for sfx in value:
+                # Don't apply title casing here in case suffix isn't one we've mapped
+                # Allows us to append back the original unaltered data
+                suffix = str(sfx)
+                if suffix.title() in _SUFFIX_VARIANTS_TO_STANDARD_SUFFIXES:
```
ericbuckley, line 200: @bamader I don't think this normalizes "jr" and "SR" to "Jr" and "Sr" respectively.  Do you think we should add logic to normalize case here, or maybe more broadly would it be best to adjust the fuzzy matcher function to normalize case before the comparisons?

bamader: Oh good catch, you're totally right. Originally, I actually had 

```
suffix = str(sfx).title()
if suffix in _SUFFIX_VARIANTS:
    suffix = _SUFFIX_VARIANTS[suffix]
normalized.append(suffix)
```

This would safely handle both of the cases you outlined (and was my approach to handling different casings but same letters in our outputs), but then I saw in one of the tests I wrote that this would apply title casing to things that weren't recognized as normalizations. So if a user had `suffix=invalid`, then in their created `Name` object, `Name.suffix` would be `Invalid` (capitalized).

I think the question is whether it's okay for us to append a "partially normalized" suffix even if we can't find a mapping. I.e., case standardization is generally considered good practice for normalization, so is it okay that if a user's suffix doesn't have an exact mapping into our standardized outputs, we still use it? My gut says this would be okay and not a problem for us to use, but that's not something I wanted to assume without other input.

What do you think?

bamader: Wanna tag @m-goggins on this too since she had the original planned vision

ericbuckley: My preference would be to not capitalize the suffix if it's not found in the mapping, mainly because that's the functionality of our state and phone number normalization processes (leave it as is if it's not a case we're prepared to handle).

bamader: Better casing logic added with new tests added based on the points you made

ericbuckley (dismissed):
🚀

ericbuckley (approved):

PR #330, opened 2025-04-25
schema.AlgorithmContext

### LINKED ISSUE #293: create algorithm_context
## Summary
Create Algorithm.algorithm_context attribute in the schema.

## Acceptance Criteria
- [x] Add "include_multiple_matches", "log_odds" and "skip_values" to algorithm_context
- [x] Add validation that log_odds are valid features and values are greater than 0
- [x] Add validation to evaluators to verify that log_odds feature is available before use
- [x] Add validation to blocking keys to verify that log_odds feature is available before use

ericbuckley:
## Description
Create AlgorithmContext nested schema object to store include_multiple_results, log_odds and skip_values.

## Related Issues
closes #293 

## Additional Notes
The `max_log_odds_points` parameter was removed from the compare method.  It's still required in the function, but now we need to retrieve the log_odds value for a feature BEFORE we call the matcher function (previously the matcher functions were retrieving this from the kwargs).  Since we're doing that in compare, it seemed simple enough to just make that a calculated value (summations are simple and quick).

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/assets/initial_algorithms.json
+++ b/src/recordlinker/assets/initial_algorithms.json
@@ -3,10 +3,23 @@
         "label": "dibbs-default",
         "description": "The core DIBBs Log-Odds Algorithm. This default, recommended algorithm uses statistical correction to adjust the links between incoming records and previously processed patients (it does so by taking advantage of the fact that some fields are more informative than others—e.g., two records matching on MRN is stronger evidence that they should be linked than if the records matched on zip code). It can be used if additional granularity in matching links is desired. However, while the DIBBs Log-Odds Algorithm can create higher-quality links, it is dependent on statistical updating and pre-calculated population analysis, which requires some work on the part of the user. For those cases where additional precision or stronger matching criteria are required, the Log-Odds algorithm is detailed below.",
         "is_default": true,
-        "include_multiple_matches": true,
+        "algorithm_context": {
+          "include_multiple_matches": true,
+          "log_odds": [
+            { "feature": "ADDRESS", "value": 8.438284928858774 },
+            { "feature": "BIRTHDATE", "value": 10.126641103800338 },
+            { "feature": "CITY", "value": 2.438553006137189 },
+            { "feature": "FIRST_NAME", "value": 6.849475906891162 },
+            { "feature": "LAST_NAME", "value": 6.350720397426025 },
+            { "feature": "IDENTIFIER", "value": 0.3051262572525359 },
+            { "feature": "SEX", "value": 0.7510419059643679 },
+            { "feature": "STATE", "value": 0.022376768992488694 },
+            { "feature": "ZIP", "value": 4.975031471124867 }
+          ],
+          "skip_values": []
+        },
         "max_missing_allowed_proportion": 0.5,
         "missing_field_points_proportion": 0.5,
-        "skip_values": [],
         "passes": [
             {
                 "label": "BLOCK_birthdate_identifier_sex_MATCH_first_name_last_name",
@@ -38,17 +51,6 @@
                         "ADDRESS": 0.9,
                         "CITY": 0.92,
                         "ZIP": 0.95
-                    },
-                    "log_odds": {
-                        "ADDRESS": 8.438284928858774,
-                        "BIRTHDATE": 10.126641103800338,
-                        "CITY": 2.438553006137189,
-                        "FIRST_NAME": 6.849475906891162,
-                        "LAST_NAME": 6.350720397426025,
-                        "IDENTIFIER:MR": 0.3051262572525359,
-                        "SEX": 0.7510419059643679,
-                        "STATE": 0.022376768992488694,
-                        "ZIP": 4.975031471124867
                     }
                 }
             },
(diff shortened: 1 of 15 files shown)


src/recordlinker/schemas/algorithm.py
```diff
@@ -44,6 +44,86 @@ def serialize_func(self, value: matchers.FeatureFunc) -> str:
...
+        if result:
+            return result
+
+        vals = value.values_to_match() if isinstance(value, Feature) else [str(value)]
+        for val in vals:
+            result = self._log_odds_mapping.get(val, None)
+            if result:
+                break
+
+        self._log_odds_cache[key] = result
```
ericbuckley, line 123: @bamader @m-goggins I could go either way on this cache key, but am starting to lean towards removing it.  The block above (117-121) is mostly just an operation of converting an object to a string and looking it up in a map (there is an edge case for identifiers).  Thoughts?

m-goggins: If I'm understanding correctly, the edge case would be having multiple identifiers whose values you can retrieve from the cache instead of recomputing?

src/recordlinker/schemas/algorithm.py
```diff
@@ -44,6 +44,86 @@ def serialize_func(self, value: matchers.FeatureFunc) -> str:
...
+    log_odds: typing.Sequence[LogOdd] = []
+    skip_values: typing.Sequence[SkipValue] = []
+
+    @pydantic.model_validator(mode="after")
+    def init_log_odds_helpers(self) -> typing.Self:
+        """
+        Initialize cache helpers for returning log odds values.
+        """
+        self._log_odds_cache: dict[str, float | None] = {}
+        self._log_odds_mapping: dict[str, float] = {str(o.feature): o.value for o in self.log_odds}
```
bamader, line 103: Is this right here basically the answer to my question earlier? Under the hood it really is a dictionary that's being accessed? I think I'm following, I just want to make sure I understand how this part of the refactor actually works because it's not something I've used before.

src/recordlinker/schemas/algorithm.py
bamader: I think I'm following. This line basically just updates the K-V pair in the caching dictionary so that if we did have to do an identifier lookup in the edge case you outlined, on future calls, we don't have to walk through that process again if the user gives us IDENTIFIER:DL but we only have IDENTIFIER. Is that right?

If that's the case, then it seems like keeping it isn't harmful, just a small quality of life thing that means we won't have to re-search the values in future calls (which I think would be likely to happen if the user doesn't update their log-odds or their algorithm, and neither of those should be changing very frequently). 

I definitely don't feel strongly, and if my thinking is correct, I may lean slightly towards keeping it but would be totally fine either way!

bamader (commented):
Sorry for the delay on getting back to this--for whatever reason it just took me a while to wrap my head around the validation layer on top of the caching layer on top of the dictionary. I think everything makes logical sense to me, just a few outstanding questions to clear up on my part!

src/recordlinker/assets/initial_algorithms.json
```diff
@@ -3,10 +3,23 @@
         "label": "dibbs-default",
         "description": "The core DIBBs Log-Odds Algorithm. This default, recommended algorithm uses statistical correction to adjust the links between incoming records and previously processed patients (it does so by taking advantage of the fact that some fields are more informative than others—e.g., two records matching on MRN is stronger evidence that they should be linked than if the records matched on zip code). It can be used if additional granularity in matching links is desired. However, while the DIBBs Log-Odds Algorithm can create higher-quality links, it is dependent on statistical updating and pre-calculated population analysis, which requires some work on the part of the user. For those cases where additional precision or stronger matching criteria are required, the Log-Odds algorithm is detailed below.",
         "is_default": true,
-        "include_multiple_matches": true,
+        "algorithm_context": {
+          "include_multiple_matches": true,
+          "log_odds": [
+            { "feature": "ADDRESS", "value": 8.438284928858774 },
```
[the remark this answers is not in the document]
ericbuckley: Both are options we could program around.  The form I have gives us a uniform data structure, so its easy to create in pydantic with type validations and auto openapi doc strings defining the allowed values.  The other way its a bit tricker to program around and we'd have to make more modifications to the way pydantic creates the openapi.json file to get the docs right.  I think the class would start off looking like the following, which doesn't give us much structure and requires us to define more rules around key typing and docs (which I'm honestly just not exactly sure how we'd do it).

```python
class LogOdd:
  __root__: Dict[str, float]
```

src/recordlinker/schemas/algorithm.py
ericbuckley: Yep, I'm using the feature/value attributes to give the schema structure that's easy to validate and document.  The `get_log_odds` method makes it quick (via caching) and easy to programmatically retrieve values.

src/recordlinker/schemas/algorithm.py
ericbuckley: Yeah it doesn't add much.  Most of the time `vals` declared on line 117 is going to be a list of 1 item.  With IDENTIFIER:XX it will be 2 items.  So in that case, the cache helps cut down on the lookups.  It's negligible though, so I'm inclined to remove it just so it reads better.

bamader: That works for me. Readability is always one of my top factors, so I'm good to remove 👍

src/recordlinker/assets/initial_algorithms.json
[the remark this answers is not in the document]
bamader: Got it, thanks for the walkthrough. Given that we'd have to modify pydantic access and openapi.json. I'm okay going with the uniform structure you have here. As long as it doesn't measurably slow anything down, I think this makes more sense in the long term to keep things coherent and tightly standardized.

bamader (dismissed):
Questions have been cleared up so giving the +1 to merge!

PR #334, opened 2025-04-28
Algorithm context advanced section

### LINKED ISSUE #296: create algorithm advanced section
## Summary
Create Algorithm.algorithm_context.advanced section for variables that we recommend not overriding.

## Acceptance Criteria
- [ ] Add "fuzzy_match_threshold", "fuzzy_match_measure", "max_missing_allowed_proportion" and "missing_field_points_proportion" to advanced section.
- [ ] Add optional "fuzzy_match_threshold" and "fuzzy_match_measure" parameters to AlgorithmPass.evaluators
- [ ] Add validation that fuzzy_match_threshold is between 0 and 1
- [ ] Add validation that fuzzy_match_measure [is valid](https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/linking/matchers.py#L20)
- [ ] All advanced values should come with defaults

ericbuckley:
## Description
Create algorithm advanced section, `algorithm.context.advanced`, to store parameters that we advise users to change with caution.  All of these parameters will now come with recommended default values.

## Related Issues
closes #296

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/linking/link.py
+++ b/src/recordlinker/linking/link.py
@@ -91,6 +91,27 @@ class LinkResult:
             self._update_score_tracking_row(earned_points, pass_lbl, rms, mmt, cmt, grade)
 
 
+def invoke(
+    evaluator: schemas.Evaluator,
+    record: schemas.PIIRecord,
+    mpi_record: schemas.PIIRecord,
+    context: schemas.AlgorithmContext,
+) -> tuple[float, bool]:
+    """
+    Invoke the evaluator function and return the result
+    """
+    fn: typing.Callable = evaluator.func.callable()
+    kwargs = {
+        "log_odds": context.get_log_odds(evaluator.feature) or 0.0,
+        "missing_field_points_proportion": context.advanced.missing_field_points_proportion,
+        "fuzzy_match_threshold": evaluator.fuzzy_match_threshold
+        or context.advanced.fuzzy_match_threshold,
+        "fuzzy_match_measure": evaluator.fuzzy_match_measure
+        or context.advanced.fuzzy_match_measure,
+    }
+    return fn(record, mpi_record, evaluator.feature, **kwargs)
+
+
 def compare(
     record: schemas.PIIRecord,
     mpi_record: schemas.PIIRecord,
@@ -114,31 +135,23 @@ def compare(
       candidate are a match, as determined by the specific matching rule
       contained in the algorithm_pass object.
     """
-    # keyword arguments to pass to comparison functions
-    kwargs: dict[typing.Any, typing.Any] = algorithm_pass.kwargs
-
     missing_field_weights: float = 0.0
     results: list[float] = []
     details: dict[str, typing.Any] = {}
     max_log_odds_points: float = 0.0
-    missing_field_propotion: float = context.advanced.missing_field_points_proportion
     max_missing_proportion: float = context.advanced.max_missing_allowed_proportion
     for evaluator in algorithm_pass.evaluators:
-        feature: schemas.Feature = evaluator.feature
-        func: typing.Callable = evaluator.func.callable()
-        log_odds: float = context.get_log_odds(feature) or 0.0
+        log_odds: float = context.get_log_odds(evaluator.feature) or 0.0
         max_log_odds_points += log_odds
         # Evaluate the comparison function, track missingness, and append the
         # score component to the list
-        result: tuple[float, bool] = func(
-            record, mpi_record, feature, log_odds, missing_field_propotion, **kwargs
-        )  # type: ignore
+        result: tuple[float, bool] = invoke(evaluator, record, mpi_record, context)
         if result[1]:
             # The field was missing, so update the running tally of how much
             # the candidate is missing overall
             missing_field_weights += log_odds
         results.append(result[0])
-        details[f"evaluator.{feature}.{func.__name__}.result"] = result
+        details[f"evaluator.{evaluator.feature}.{evaluator.func}.result"] = result
 
     # Make sure this score wasn't just accumulated with missing checks
     if missing_field_weights <= (max_missing_proportion * max_log_odds_points):
--- a/src/recordlinker/linking/link.py
+++ b/src/recordlinker/linking/link.py
@@ -94,11 +94,10 @@ class LinkResult:
 def compare(
     record: schemas.PIIRecord,
     mpi_record: schemas.PIIRecord,
-    max_log_odds_points: float,
     max_allowed_missingness_proportion: float,
     missing_field_points_proportion: float,
     algorithm_pass: schemas.AlgorithmPass,
-    log_odds_weights: dict[str, float],
+    context: schemas.AlgorithmContext,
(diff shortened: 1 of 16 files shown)


src/recordlinker/linking/link.py
```diff
@@ -91,11 +91,30 @@ def check_and_update_score(self, earned_points, pass_lbl, rms, mmt, cmt, grade):
             self._update_score_tracking_row(earned_points, pass_lbl, rms, mmt, cmt, grade)
 
 
+def invoke(
```
bamader, line 94: Could we rename this something like `invoke_evaluator` or `apply_comparison` or just something like that (just giving examples, I'm good with any combination that combines some application verb with a noun suggesting evaluation or comparison)? No logical problems in the code, but when looking at the function on its own outside of line 148, its context isn't super clear.

src/recordlinker/linking/matchers.py
```diff
@@ -55,40 +55,7 @@ class AvailableKwarg(enum.Enum):
...
-
-    threshold: float = 0.7
-    if "thresholds" in kwargs:
-        if col in kwargs["thresholds"]:
-            threshold = kwargs["thresholds"][col]
-    elif "threshold" in kwargs:
-        threshold = kwargs["threshold"]
-
-    return (similarity_measure, threshold)
+    pass
```
bamader, line 58: If the only code affiliated with the class is just `pass`, do we still need to keep this structure around? Seems like something we could probably just remove and re-type throughout the rest of the codebase.

src/recordlinker/linking/matchers.py
```diff
@@ -144,6 +111,8 @@ def compare_probabilistic_fuzzy_match(
     key: Feature,
     log_odds: float,
     missing_field_points_proportion: float,
+    fuzzy_match_measure: SIMILARITY_MEASURES,
+    fuzzy_match_threshold: float,
```
bamader, line 115: While this doesn't break anything, prior to this change we had a uniformity in our `compare_` functions. Both `compare_probabilistic_exact` and `compare_probabilistic_fuzzy` had the same function signature, which in my opinion helped presented a unified parameter input and uniform standard to users when looking at what exactly a feature function was. It helped standardize our own code definitions, documentation, as well as user expectations. I do think the need for some of that standardization has dropped over time, since we now support only the fuzzy matching functions rather than other kinds of comparison matchers, but just pointing it out since uniformity was a good secondary benefit of that design structure. This isn't a reason not to merge, but if there's a way we can preserve those identical signatures, I think it would be a good thing to do so. Just my two cents though.

src/recordlinker/linking/matchers.py
```diff
@@ -177,15 +148,14 @@ def compare_probabilistic_fuzzy_match(
         # Return early if a field is missing, and log that was the case
         return (missing_field_points_proportion * log_odds, True)
 
-    similarity_measure, threshold = _get_fuzzy_params(str(key.attribute), **kwargs)
```
bamader, line 180: Just adding on to the above comment, the uniformity is why we had this `get_params` functions which didn't do a whole lot else. We don't necessarily need to use this, though, just adding some context.

bamader (commented):
Logic all looks good! Just some comments around names and possible parameter standardization, but they're small things and this is just about there.

src/recordlinker/linking/matchers.py
ericbuckley: We don't.  I was planning on cleaning this up along with a few other things (like kwargs in initial_algorithms) in #344.

bamader: Ah perfect, that makes total sense. Feel free to ignore here!

src/recordlinker/linking/matchers.py
ericbuckley: The signatures of the `compare_` methods in main are the same, however its a little misleading for two reasons.

1. `**kwargs` is hiding any differences in parameters sent.  In main, its true that its always sent the same value for a pass, but the compare_exact method ignores those values.
2. `_get_fuzzy_params` provides default values for the measure and thresholds if they are missing.  This is something I wanted to surface to the user (via the advanced fuzzy_match_* parameters) so they know whats happening in case they don't override the defaults.  As is, users only see the "JaroWinkler / 0.7" default values if they read the code.

I did try a version of leaving the signatures the same between compare_fuzzy and compare_exact, but added some assertions to the beginning of the compare_fuzzy method to verify that they were specified in the kwargs.  At first, I didn't like that version because mypy isn't able to catch errors in which the fuzzy measure and threshold values are missing, when they are function parameters mypy can.

However, given we call that method dynamically (via `evaluator.func.callable()(....)`), mypy isn't able to run type checks.  Its really only able to do it for test cases, which has limited value.

I'm ok going back to that version of the calls, but I'd still want to add the assertions to the compare_fuzzy method to verify that they were specified in the kwargs.

ericbuckley: For reference, this is what that would look like.  Let me know what you think.

https://github.com/CDCgov/RecordLinker/pull/334/commits/fcc9e2252ec94003e1ff2f1d2a77dd3eb091f679

src/recordlinker/linking/link.py
ericbuckley: changed to `invoke_evaluator`.

src/recordlinker/linking/matchers.py
bamader: Thanks for putting up this comparison. Very helpful to see what things look like. After checking it out, I think I agree with your original implementation. You're totally right that the kwargs were hiding any parameter differences, and I hadn't considered the point about surfacing default behavior to the user regarding threshold values. That's really valuable. I also think the compare_fuzzy method gets a little cluttered and less readable with these assertions, where your other version was very clean and straightforward. If it's okay with you, I think we should go back to your first version with the different signatures. It just feels more straightforward, and I don't really think we'll be implementing new comparison methods in the future to make preserving legacy signatures a meaningful point of standardization.

Also, I'm super glad you surfaced the default threshold point because I totally forgot that was in there. We actually want to change that to 0.9! We scrapped the baseline JaroWinkler of 0.7 during prod testing with LAC because we found it was way too permissive and loose, but clearly we missed this buried 0.7 for exactly the thing your change is addressing. All the more reason to clean up the signatures!

src/recordlinker/linking/matchers.py
bamader: (now irrelevant, gonna resolve this because the first implementation was cleaner)

src/recordlinker/linking/matchers.py
ericbuckley: Cool, I'll do that.  And yes, the [default threshold will be 0.9](https://github.com/CDCgov/RecordLinker/pull/334/files#diff-f8b75e7f2665e1eea4eee28b690b98fa9fd3b2687327a847c80df04145ecb069R101).

bamader (approved):
Looks good--thanks for the back and forth on this! This cleanup refactor is really starting to come together 💪

PR #335, opened 2025-04-28
Add link and unlink endpoints

### LINKED ISSUE #313: record link/unlink endpoints
## Summary
Create two new endpoints, `POST /api/demo/record/<id:int>/link` and `POST /api/demo/record/<id:int>/unlink` to capture state changes the user is making on the given record.

## Acceptance Criteria
- [ ] Implement the link endpoint
- [ ] Implement the unlink endpoint
- [ ] Update the list record API endpoint with state data
- [ ] Update the get record API endpoint with state data
- [ ] Test cases

## Details / Tasks
Both endpoints should be store info into the session to capture that the record in question has either linked or unlinked the record.  When the list records or get record API endpoints are invoked, this linked vs unlinked information should be included in the response.  By default, all records will return the payload `linked: null` when retrieved from the `GET /api/demo/record` or `GET /api/demo/record/<id:int>` endpoints, but if a value is captured in the session, we should update the `linked` property to "true" or "false".

m-goggins:
## Description
This PR adds 2 new endpoints: 
1) "link" endpoint to update the `linked_status` to `True` when a user clicks on the "link" button. It also updates the `person_id` of the incoming record to match the `person_id` of the potential match record. The data is also stored in the session.
2) "unlink" endpoint to update the `linked_status` to `False` when a user clicks on the "unlink" button. It also updates the `person_id` of the incoming record to `null`. The data is also stored in the session.

For the data that is stored in the cookie session, I'm not sure if it is better just to save the boolean value of the linked status (what I've done) or if all the data is better. It seems like the values tend to be short but I'm looking for feedback/best practices! 

## Related Issues
Closes #313 

## Additional Notes
I also removed the `max_age` of a cookie since we will be handling the reset in a follow up ticket.

### DIFF
--- a/src/api/recordlinker/routes/demo_router.py
+++ b/src/api/recordlinker/routes/demo_router.py
@@ -105,7 +105,7 @@ def link_match(
     session_store.save_session(
         response,
         key="linked_status",
-        data=match_review_record,
+        data=match_review_record["linked"],
     )
     return match_review_record
 
@@ -137,6 +137,6 @@ def unlink_match(
     session_store.save_session(
         response,
         key="linked_status",
-        data=match_review_record,
+        data=match_review_record["linked"],
     )
     return match_review_record
--- a/src/api/recordlinker/routes/demo_router.py
+++ b/src/api/recordlinker/routes/demo_router.py
@@ -6,6 +6,7 @@ This module implements the demo router for the RecordLinker API. Exposing
 the API endpoints for the demo UI.
 """
 
+import copy
 import typing
 
 import fastapi
@@ -59,7 +60,6 @@ def update_session_linked_status(
         )
         or {}
     )
-
     # Update the session data & save
     d.update({patient_reference_id: linked_status})
     session_store.save_session(
@@ -85,15 +85,15 @@ def get_demo_data(
         request,
         key="linked_status",
     )
-
+    d = copy.deepcopy(data)  # copy to avoid modifying the original data
     if linked_status:
-        for record in data:
+        for record in d:
             patient_id = record["incoming_record"]["patient_id"]
             if str(patient_id) in linked_status:
                 record["linked"] = linked_status[str(patient_id)]
 
     # Filter and sort the data based on the linked status
-    filtered_sorted = filter_and_sort(data, status)
+    filtered_sorted = filter_and_sort(d, status)
 
     return filtered_sorted
 
@@ -115,15 +115,19 @@ def get_match_review_records(
     )
     if match_review_record is None:
         raise fastapi.HTTPException(status_code=fastapi.status.HTTP_404_NOT_FOUND)
+    # Copy the record to avoid modifying the original data
+    record_copy = copy.deepcopy(match_review_record)
 
     # Update the linked status based on the session store
     linked_status = session_store.load_session(
         request,
         key="linked_status",
     )
+
     if linked_status and str(patient_reference_id) in linked_status:
-        match_review_record["linked"] = linked_status[str(patient_reference_id)]
-    return match_review_record
+        record_copy["linked"] = linked_status[str(patient_reference_id)]
+
+    return record_copy
 
 
 @router.post(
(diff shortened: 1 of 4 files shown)


ericbuckley:
@m-goggins I saw your additional notes about testing the cookie.   You can also use your browser devtools to manually delete a cookie.

Screenshot 2025-04-29 at 08 19 36

src/api/recordlinker/routes/demo_router.py
```diff
@@ -73,3 +74,69 @@ def get_match_review_records(
...
+    # Update the incoming record with the person_id from the potential match
+    match_review_record["incoming_record"]["person_id"] = match_review_record["potential_match"][0][
+        "person_id"
+    ]
+
+    # Save session (modifies the real response)
+    session_store.save_session(
+        response,
+        key="linked_status",
+        data=match_review_record["linked"],
```
ericbuckley, line 108: I believe this is going to overwrite any previous cookie value.  For example, if the user marks demo record 1 as linked, then marks demo record 2 as unlinked, the second operation will overwrite the previous data stored on record 1.

src/api/recordlinker/routes/demo_router.py
```diff
@@ -73,3 +74,69 @@ def get_match_review_records(
     if match_review_record is None:
         raise fastapi.HTTPException(status_code=fastapi.status.HTTP_404_NOT_FOUND)
     return match_review_record
+
```
ericbuckley, line 77: We'll need to update the get_demo_data and get_match_review_records methods to use the cookie data as well.  When the user goes back to the listing page, we want to properly show/filter based on what's been linked/unlinked.

src/api/recordlinker/routes/demo_router.py
m-goggins: Ah you're right. I previously had the key set to `linked_status_{patient_reference_id}` to avoid this but it was flagged as a potential issue because users could inject harmful code. However, the endpoint is typed to only handle integers as input so maybe that prevents inject issues. What do you think?

m-goggins: Another way I could do it, is to keep the key as is, but change the data to be a dictionary with patient_reference_id as the key and True/False as the value like:

```
key="linked_status",
data={"1":True, "3":False}
```

m-goggins: I'm liking the 2nd option better because all data is stored in a single cookie, which will make it easier to clear when we implement the reset button. And it's easier to retrieve all the data from a single cookie to render the data on the queue page.

johanna-skylight: Agree. It makes sense for the session to be created when the user hits the `/GET /api/demo/record` endpoint for the first time. At this stage a copy of the json is created for the specific session/user and any subsequent calls to the demo endpoints retrieve, modify and update the record object in the session.

TL;DR; Line 103 should obtain the record from the session and modify it then save the updated object back. The session will contain a record at this point because it gets created the first time the user calls the `/GET /api/demo/record` endpoint.

johanna-skylight (commented):
Great work Marcelle! I have left a few comments that I think align with Eric's comments. I'm not a python expert so LMK if anything needs more clarification. Thanks!

src/api/recordlinker/routes/demo_router.py
m-goggins: I modified it so that link/unlink update the linked status in the cookie (not the underlying data) and the GET `/api/demo/record` now checks the cookie for updates before returning the data to reflect any changes that were made by someone clicking link or unlink. Does that make sense?

src/api/recordlinker/session_store.py
```diff
@@ -14,7 +14,6 @@ def save_session(response: fastapi.Response, key: str, data: dict, **kwargs: typ
     response.set_cookie(
         key,
         value=serializer.dumps(data),
-        max_age=0,
```
ericbuckley, line 17: Ah good catch, max_age of 0 deletes it.

ericbuckley (approved):
Looks great, thanks @m-goggins

PR #346, opened 2025-05-01
Handle suffix during evaluation

### LINKED ISSUE #299: Implement FULL_FIRST_NAME FeatureAttribute
## Summary
Implement a new feature attribute called FULL_FIRST_NAME that prepends normalized suffixes (where applicable) to first names during evaluation so that suffixes may be included.

## Acceptance Criteria
- [ ] FULL_FIRST_NAME compares first names, including normalized suffixes (where applicable); see task details about why we should only include normalized suffixes
- [ ] The implementation in feature_iter includes a check that suffixed has been normalized before prepending it to the first name
- [ ] When a record does not have a suffix, FULL_FIRST_NAME includes just the first name
- [ ] Tests are added to demonstrate multiple edge cases, such as comparing two names with normalized suffixes, at least 1 first name has a suffix that is not normalized (and ignored), at least 1 name does not have a suffix, etc.
- [ ] Documentation is updated to reflect the new feature attribute

## Details / Tasks
We should only prepend **normalized** suffixes. If there is text in the suffix field but it could not be normalized, we should not prepend it. For example, if the suffix is "the 2nd" and first name is "John", FULL_FIRST_NAME for this record should be "JOHN", not "THE2NDJOHN". This would likely lead to very low similarity scores for first name due to bad suffix normalization, which we do not want.

bamader:
## Description
This PR adds the final piece of handling suffixes during name evaluation. Previously, we used them only during blocking, but this change set now includes handling during evaluation as well. In order to still maintain the First_Name field as its own readable entity, we've simply re-used the `should-prepend` parameter in `feature_iter`. That way, it's only during the very specific cases we've outlined that we combine a name and a suffix.

## Related Issues
Closes #299 

## Checklist

## Checklist for Reviewers

### DIFF
--- a/tests/unit/linking/test_link.py
+++ b/tests/unit/linking/test_link.py
@@ -529,6 +529,115 @@ class TestLinkRecordAgainstMpi:
         # low-value missing field, but fails user's overall completeness constraint
         assert matches == [False, False]
 
+    def test_no_match_one_suffix_one_not(
+        self,
+        session,
+        default_algorithm,
+        patients: list[schemas.PIIRecord]
+    ):
+        # Make a deep copy of the first patient, give it a suffix
+        patients = [patients[0]]
+        duplicate = copy.deepcopy(patients[0])
+        duplicate.external_id = str(uuid.uuid4())
+        duplicate.name[0].suffix = ["Sr"]
+        patients.append(duplicate)
+
+        matches: list[bool] = []
+        mapped_patients: dict[str, int] = collections.defaultdict(int)
+        for data in patients[:2]:
+            (_, person, results, _) = link.link_record_against_mpi(data, session, default_algorithm)
+            matches.append(bool(person and results))
+            mapped_patients[person.reference_id] += 1
+
+        # First patient inserted into empty MPI, no match
+        # Second patient fails eval in first pass and blocking in second
+        # due to suffix
+        assert matches == [False, False]
+
+    def test_no_match_same_name_diff_suffixes(
+        self,
+        session,
+        default_algorithm,
+        patients: list[schemas.PIIRecord]
+    ):
+        '''
+        NOTE: This catches the Jr/Sr edge case of a parent and child
+        living at the same address with the same first name. It even
+        goes a step farther and gives them the same birthday.
+        '''
+        # Give both patients the same name before duplication, change
+        # suffixes after
+        patients = [patients[0]]
+        duplicate = copy.deepcopy(patients[0])
+        patients[0].name[0].suffix = ["Sr"]
+        duplicate.external_id = str(uuid.uuid4())
+        duplicate.name[0].suffix = ["Jr"]
+        patients.append(duplicate)
+
+        matches: list[bool] = []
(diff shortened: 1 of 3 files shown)


tests/unit/linking/test_link.py
```diff
@@ -529,6 +529,115 @@ def test_missing_field_points_exceed_max_missingness_fraction(
...
+    def test_no_match_same_name_diff_suffixes(
+        self,
+        session,
+        default_algorithm,
+        patients: list[schemas.PIIRecord]
+    ):
+        '''
+        NOTE: This catches the Jr/Sr edge case of a parent and child
+        living at the same address with the same first name. It even
+        goes a step farther and gives them the same birthday.
```
m-goggins, line 566: dang, incredible!

tests/unit/linking/test_link.py
```diff
@@ -529,6 +529,115 @@ def test_missing_field_points_exceed_max_missingness_fraction(
...
+        mapped_patients: dict[str, int] = collections.defaultdict(int)
+        for data in patients[:2]:
+            (_, person, results, _) = link.link_record_against_mpi(data, session, default_algorithm)
+            matches.append(bool(person and results))
+            mapped_patients[person.reference_id] += 1
+
+        # First patient inserted into empty MPI, no match
+        # Second patient fails eval in first pass and blocking in second
+        # due to suffix
+        assert matches == [False, False]
```
m-goggins, line 555: This is the edge case I was interested and was the only hesitation I had with implementing the suffix evaluation for `FIRST_NAME` as opposed to keep it as is and introducing `FULL_FIRST_NAME`. I'll make a follow up ticket up adding Jr/Sr to the edge case appendix with a specific call out for this case.

m-goggins (approved):
Looks great!

bamader:
Ran the algorithm match tests just to see, and we caught one more true positive because we now handle Jr/Sr edge cases: 

```
Results:
32.0 true positives correctly identified
15.0 true negatives correctly identified
2.0 false positives misidentified
21.0 false negatives misidentified

Sensitivity: 0.6037735849056604
Specificity: 0.8823529411764706
F1-Score: 0.735632183908046
PPV: 0.9411764705882353

Possible Matches, NBS Perspective
3.0 NBS-labelled possible matches we graded 'certain'
7.0 NBS-labelled possible matches we graded 'certainly-not'
0.0 NBS-labelled possible matches we graded 'possible-match'

Possible Matches, DIBBs Perspective
0.0 DIBBs-labelled possible matches NBS graded 'certain'
0.0 DIBBs-labelled possible matches NBS graded 'certainly-not'
0.0 DIBBs-labelled possible matches NBS graded 'possible-match'
```

tests/unit/linking/test_link.py
bamader: Prepending suffix was so helpful. The difference in J vs S makes it auto-fail blocking in pass 2, so address and birthdate never have a chance to come up, and in pass 1, the letter difference at the first character really tanks JaroWinkler from the start because earlier typos are more heavily weighted. It was a great idea on your part!

Issue #317, opened 2025-04-17
Integrate link and unlink action (with toast confirmation)


johanna-skylight:
## Summary
A user should be able to mark a record as link or do not link. After the action has been performed a toast will display to the user confirming that the change has been successful.

## Acceptance Criteria
- [ ] Link and do not link buttons successfully mark the record as reviewed.
- [ ] A toast will be presented to the user confirming the action was successful. 
- [ ] The user should be taken back to the record queue page automatically..

## Details / Tasks
List the subtasks or steps needed to complete this implementation.

1. Integrate the link / do not link buttons with the backend endpoint that marks the record as reviewed.
Endpoint link record: POST /api/demo/record/<id:int>/link
Endpoint unlink record: POST /api/demo/record/<id:int>/unlink

2. Integrate the toast dependency that query connector is using. 
3. Implement redirect to the queue page after the record state gets updated.
4. Display confirmation toast after the record state gets updated.

PR #351, opened 2025-05-05
Integrate link unlink endpoints

johanna-skylight:
## Description
This PR implements the link/unlink endpoints and the toast component. This work includes:
- Includes react-toastify dependency. Same library used in query connector.
- Integrates the toast component in the `root layout`.
- Implements a wrapper method to trigger the toast that takes care of the component customization with the uswds alert component.
- Clean up of the page-container classes. New implementation translates to less css lines after compilation.
- Adds styles in global for toastify classes.

## Related Issues
Implements #317 

## Additional Notes
Screenshot 2025-05-05 at 1 15 27 PM

### DIFF
--- a/src/ui/src/data/matchReview.ts
+++ b/src/ui/src/data/matchReview.ts
@@ -42,6 +42,7 @@ export async function unlinkRecordAndMatch(
 ): Promise<RecordMatch> {
   const response = await fetch(`${API_URL}/demo/record/${id}/unlink`, {
     method: "POST",
+   credentials: "include",
   });
 
   if (response.ok) {
--- a/src/ui/src/data/matchReview.ts
+++ b/src/ui/src/data/matchReview.ts
@@ -23,6 +23,7 @@ export async function linkRecordAndMatch(
 ): Promise<RecordMatch> {
   const response = await fetch(`${API_URL}/demo/record/${id}/link`, {
     method: "POST",
+     credentials: "include",
   });
 
   if (response.ok) {
--- a/src/ui/src/data/matchReview.ts
+++ b/src/ui/src/data/matchReview.ts
@@ -17,3 +17,41 @@ export async function getRecordMatch(id: string | null): Promise<RecordMatch> {
     );
   }
 }
+
+export async function linkRecordAndMatch(
+  id: string | null,
+): Promise<RecordMatch> {
+  const response = await fetch(`${API_URL}/demo/record/${id}/link`, {
+    method: "POST",
+  });
+
+  if (response.ok) {
+    const serializedRecordMatch = await response.json();
+    return deserializeRecordMatch(serializedRecordMatch);
+  } else {
+    throw new AppError(
+      "linkRecordAndMatch",
+      "unsuccessful HTTP response",
+      response.status,
+    );
+  }
+}
+
+export async function unlinkRecordAndMatch(
+  id: string | null,
+): Promise<RecordMatch> {
+  const response = await fetch(`${API_URL}/demo/record/${id}/unlink`, {
+    method: "POST",
+  });
+
+  if (response.ok) {
+    const serializedRecordMatch = await response.json();
+    return deserializeRecordMatch(serializedRecordMatch);
+  } else {
+    throw new AppError(
+      "unlinkRecordAndMatch",
+      "unsuccessful HTTP response",
+      response.status,
+    );
+  }
+}
--- a/compose.yml
+++ b/compose.yml
@@ -36,6 +36,7 @@ services:
         - "8080:8080"
     environment:
       DB_URI: "postgresql+psycopg2://postgres:pw@db:5432/postgres"
+      SECRET_KEY: "not_a_good_secret_key"
       TEST_DB_URI: "sqlite:///:memory:"
       OTEL_TRACES_EXPORTER: "otlp"
       OTEL_EXPORTER_OTLP_ENDPOINT: "http://jaeger:4317"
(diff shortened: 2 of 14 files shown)


src/ui/src/data/matchReview.ts
```diff
@@ -17,3 +17,41 @@ export async function getRecordMatch(id: string | null): Promise<RecordMatch> {
     );
   }
 }
+
+export async function linkRecordAndMatch(
+  id: string | null,
```
ericbuckley, line 22: Curious as to in what case the id will be null?  My thought is if the id is null, we'd expect an API error to be raised, so maybe we should raise a different error upstream from this call.

ericbuckley (approved):
Looks good, just 1 question on error handling.

src/ui/src/data/matchReview.ts
johanna-skylight: Just because those types match the return types of the method that grabs the id from the url. I agree with `if the id is null, we'd expect an API error to be raised` and that is the expected behavior. The frontend client will still make the API call (HTTP request) with null as the ID. 

The client is expecting that the endpoint will return an error 400. That error response then will be passed to the page and the server error screen will display.

Let me know if what you are thinking is that the id as null is handled at the client level at which point we would not even start the HTTP request.

johanna-skylight: example when id is null.
Screenshot 2025-05-05 at 2 49 38 PM

ericbuckley: A `POST /api/demo/record/null/link` will return a 422, not a 400.  Will it still be caught?

I see your reasoning here, since it's just passing on data from the original API that returned the id, if its null, its still an API error because the original GET call returned a value of null.  I guess its just an optimization in that case.

johanna-skylight: yes, at this point all http errors (except the 404) will be handled in the page by presenting the server error screen to the user. The scenario of  calling `POST /api/demo/record/null/link` with a null id is not possible because the server error screen will not display the link and unlink buttons to the user. 

In another scenario. If the there is an error response from the POST /api/demo/record/[id]/link endpoint then the error is handled by the page and presented to the user as a toast notification. This error could be possible because we have issues during serialization or because the API server had some other issue while processing the request, but null id will not be the cause.

johanna-skylight: @ericbuckley it could also be that a better experience would be to present the user with a different screen that instructs to go to the match queue page because a record was not properly selected, instead of displaying it as a server error.

We could create a ticket to follow up on this. LMK!

src/ui/src/data/matchReview.ts
```diff
@@ -17,3 +17,41 @@ export async function getRecordMatch(id: string | null): Promise<RecordMatch> {
...
+): Promise<RecordMatch> {
+  const response = await fetch(`${API_URL}/demo/record/${id}/link`, {
+    method: "POST",
+  });
```
m-goggins, line 26:
```suggestion
    method: "POST",
     credentials: "include",
  });
```

src/ui/src/data/matchReview.ts
```diff
@@ -17,3 +17,41 @@ export async function getRecordMatch(id: string | null): Promise<RecordMatch> {
...
+      response.status,
+    );
+  }
+}
+
+export async function unlinkRecordAndMatch(
+  id: string | null,
+): Promise<RecordMatch> {
+  const response = await fetch(`${API_URL}/demo/record/${id}/unlink`, {
+    method: "POST",
```
m-goggins, line 44:
```suggestion
  const response = await fetch(`${API_URL}/demo/record/${id}/unlink`, {
    method: "POST",
   credentials: "include",
```

m-goggins (changes requested):
Looks good to me! Just need to make sure we are including credentials so that the same cookie gets passed around (instead of creating a new one). I only found this out while testing the reset button.

src/ui/src/data/matchReview.ts
ericbuckley: I'm guessing this is necessary because in dev we are using 2 hosts, and the default samesite value doesn't work.  I do wonder about this though, as I see were setting the cookie samesite value to "lax".  But if it works in dev, then I guess lets just go with it, it shouldn't be an issue at all in demo site since everything will come from recordlinker.dibbs.demo

> Note that if a cookie's [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#samesitesamesite-value) attribute is set to Strict or Lax, then the cookie will not be sent cross-site, even if credentials is set to include.

src/ui/src/data/matchReview.ts
ericbuckley: I think its fine as a server error, this shouldn't ever happen and if it does, its likely because our API server is down for some reason.

src/ui/src/data/matchReview.ts
johanna-skylight: Thanks for catching this!

src/ui/src/data/matchReview.ts
m-goggins: I was going to make the recommendation that we think about moving to use React Query (especially Mutation) to help us handle these fetches. I'm less familiar with Next JS than Johanna but it seems to handle credentials (among other things that we'd want handled automatically).

m-goggins (approved):

Issue #309, opened 2025-04-16
Update content of the record review page


johanna-skylight:
## Summary
The record review page was created with placeholder content. This ticket updates it with the official copy.

## Acceptance Criteria
- [ ] Record review page displays official copy.
- [ ]  Url is update if applicable.

## Details / Tasks
List the subtasks or steps needed to complete this implementation.
1. Update content of the page with official copy.
2. Update urls and links to the page where necessary.

Issue #304, opened 2025-04-16
Update content of the record queue page


johanna-skylight:
## Summary
The record queue page was created with placeholder content. This ticket updates it with the official copy.

## Acceptance Criteria
- [ ] Record queue page displays official copy.
- [ ] Url is update if applicable.

## Details / Tasks
List the subtasks or steps needed to complete this implementation.

1. Update content of the page with official copy.
2. Update urls and links to the page where necessary.

Issue #300, opened 2025-04-16
Update landing page content


johanna-skylight:
## Summary
Currently the landing page contains placeholder copy. The team needs to update it once the copy has been finalized and approved.

## Acceptance Criteria
- [ ] Page content has been updated.
- [ ] Diagram has been updated.

## Details / Tasks
List the subtasks or steps needed to complete this implementation.

1. Update the copy of the landing page.
2. Update the diagram in the page. [Please compress ](https://www.freeconvert.com/image-compressor)the image before introducing it in the app.
3. If applicable update the current `alt` text of the diagram.

PR #353, opened 2025-05-06
Updates copy and implements last round of clean ups

johanna-skylight:
## Description
This PR updates the copy on all pages and makes one last past of fixing styles.
- Copy and diagrams updated across pages.
- Styles in the landing page were updated to match latest figma.
- Added toast to display error message to user if resetting the demo fails.
- Added code in the `match review page` to handle 422 code error (record id null) with empty fallback component with  `invalid record` message. The server error screen started to felt too much after the convo with @ericbuckley in my previous PR.
- Implemented `CaptionImage` component.
- Implemented http `get, post` util methods. I looked into the `credentials: include` configuration a little bit more. It is correct that this is only needed due to our local setup that requires allowing cross origin. I did two things while implementing the util methods: 
- 1) Made sure that the `credentials: include` is set in requests only during local development.
- 2) I condensed boilerplate code across the data layer files that included default values and access to the `API_URL` constant. 

Still the team needs to decide if it will add a dependency to make http request but this makes things cleaner in the meantime.

## Related Issues
Implements #300 , #304 , #309 

## Additional Notes
Too many screens for screenshots! sorry.


src/ui/src/app/page.tsx
```diff
@@ -55,80 +56,112 @@ const Home: React.FC = () => {
...
+        />
+        <h2 className="margin-top-8">Where we&apos;re going </h2>
+        <p className="margin-top-1 margin-bottom-4">
+          Record Linker is currently at the product validation stage. Our team
+          is working with users at several public health departments to develop
+          a standalone Master Patient Index that enables patient matching across
+          internal tools (e.g., disease surveillance, vital records, and
+          immunization information systems), thereby improving interoperability
+          between these datasets.
+        </p>
```
m-goggins, line 156: I know we got the announcement during retro (after you put up this PR) but I'm wondering if we should modify this again.

m-goggins (approved):
Approving but think we should update the "where are we going section" ahead of CSTE in a follow up round of content edits.

src/ui/src/app/page.tsx
johanna-skylight: Good catch!

johanna-skylight:
@m-goggins @ericbuckley I will merge this for now and will address the copy updates in a separate PR

PR #360, opened 2025-05-07
Fix bug in scripts/gen_seed_data.py

### LINKED ISSUE #359: Fix bug in `gen_seed_data` script
Found a bug in the way we were generating Race data using Faker that probably cropped up during the recent changes we've made to `PIIRecord`. I thought this script might be useful for jurisdictions like WA, who may want to generate their own data for testing RL.

m-goggins:
## Description
Somewhere in our updates to `PIIRecord`, we change the expectation for Race to a string. Faker produces a Race `object`, which I changed to a string so that we can continue to run `gen_seed_data.py`. 

Original error:
```
pii.py", line 411, in parse_race
    return [Race.parse(v) for v in value]
                                   ^^^^^
TypeError: 'Race' object is not iterable
```

## Related Issues
Fixes #359 

## Additional Notes
I'm guessing this script will be useful for jurisdictions testing out Record Linker so I wanted to make sure it stayed current.

### DIFF
--- a/scripts/gen_seed_test_data.py
+++ b/scripts/gen_seed_test_data.py
@@ -75,7 +75,7 @@ def _generate_random_pii_record(faker):
                 use=random.choice(["home", "work", "mobile"]),
             )
         ],
-        race=random.choice(list(Race)),
+        race=str(random.choice(list(Race))),
         identifiers=list(_generate_random_identifiers(random.randint(1, 3), faker)),
     )
 
--- a/scripts/gen_seed_test_data.py
+++ b/scripts/gen_seed_test_data.py
@@ -75,7 +75,7 @@ def _generate_random_pii_record(faker):
                 use=random.choice(["home", "work", "mobile"]),
             )
         ],
-        race=str(random.choice(list(Race))),
+        race=[str(random.choice(list(Race)))],
         identifiers=list(_generate_random_identifiers(random.randint(1, 3), faker)),
     )
 


scripts/gen_seed_test_data.py
```diff
@@ -75,7 +75,7 @@ def _generate_random_pii_record(faker):
                 use=random.choice(["home", "work", "mobile"]),
             )
         ],
-        race=random.choice(list(Race)),
+        race=str(random.choice(list(Race))),
```
ericbuckley, line 78: Right now, the `race` field accepts a list of race strings and not a single one.  We made this change a while back as multiple race values are supported in FHIR.  I'm interested though in what behavior you observed before making this change?

https://github.com/CDCgov/RecordLinker/blob/main/src/recordlinker/schemas/pii.py#L366

m-goggins: Ah, definitely should have included the error in the PR:
```
pii.py", line 411, in parse_race
    return [Race.parse(v) for v in value]
                                   ^^^^^
TypeError: 'Race' object is not iterable
```

m-goggins: Are you seeing any errors when you run the script?

ericbuckley: When I run the change, it shows up as a list of OTHER values.  For example

```json
"race": [
  "OTHER",
  "OTHER",
  "OTHER",
  "OTHER",
  "OTHER",
]
```

I think we might need to wrap it in a list.  Otherwise line 411 in pii.py is just iterating over the string and processing each character as a race value, all of which get categorized as OTHER.

ericbuckley (approved):

PR #364, opened 2025-05-08
Expand and scramble test cases

### LINKED ISSUE #354: Expand NBS test set

bamader:
## Description
This PR adds a script that allow us to combinatorially expand and scramble the test cases NBS provided for us. We leave each of the original test cases unchanged, but we generate a random number of duplicates of each case to mess with. We then dropout some fields to simulate missingness, as well as apply random likelihoods of edit distances to string values in algorithm-relevant fields. All of this is controlled by parameters at the top of the script (which only needs to be run once to make the data-set, doesn't need to be run each time we run the algorithm tests). Of note, the script right now copies the match decisions of the original test case into each duplicate we generate from it, no matter how much we then scramble and mangle the duplicate case. This is the reason we have poor performance on this set, so that's expected. Before sharing this out, we should consider "re-grading" our expanded set to have better labels of when things should and shouldn't match. Also, since the script is customizable, we can also generate different test expansion cases if we want to simulate "pretty close" data (low randomness scrambling) vs "really bad data" (lots of scrambling). We'd need to grade each separately, but this could be useful for purposes of showing performance in different contexts.

## Related Issues
Closes #354 

## Checklist

## Checklist for Reviewers

### DIFF
--- a/tests/algorithm/scripts/expand_test_data.py
+++ b/tests/algorithm/scripts/expand_test_data.py
@@ -1,0 +1,243 @@
+import copy
+import random
+import re
+import string
+
+BASE_TEST_FILE = "../data/nbs_test.csv"
+EXPANDED_OUTPUT_FILE = "../data/expanded_nbs_test.csv"
+
+ALGORITHM_RELEVANT_COLUMNS = [
+    "BIRTHDATE",
+    "FIRST",
+    "LAST",
+    "SUFFIX",
+    "GENDER",
+    "ADDRESS",
+    "CITY",
+    "ZIP",
+    "SSN"
+]
+
+# Range of the number of copies to make of each test case, before
+# applying scrambling. Value is chosen randomly within the range.
+# Note that the range of how many cases you end up with for each
+# starting case is this range + 1, since you have the original case
+# plus the duplicates (we want to preserve the original test, too)
+MIN_DUPLICATE_CASES = 2
+MAX_DUPLICATE_CASES = 6
+
+# Controls for the random dropout of field information to simulate
+# missingness issues in incoming data. For each duplicate created
+# for a test case, there is a desired probability to apply dropout
+# to some number of fields in that duplicate. If that duplicate is
+# randomly chosen to experience dropout, then a random number of
+# fields in the range specified are blanked out.
+CHANCE_TO_DROP_FIELDS = 0.5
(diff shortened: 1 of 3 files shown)


bamader:
Performance of the algorithm tests on this initial run

Screenshot 2025-05-08 at 2 48 25 PM

tests/algorithm/scripts/expand_test_data.py
```diff
@@ -0,0 +1,243 @@
...
+    data that might be relevant to the algorithm.
+    """
+    missing_fields = []
+    for field in ALGORITHM_RELEVANT_COLUMNS:
+        if test_case[field] is None or test_case[field] == "":
+            missing_fields.append(field)
+    return missing_fields
+
+
+def apply_field_scrambling(
```
ericbuckley, line 65: I think this method could be really useful for some concepts I've been thinking about for load testing.  The big difference is we'll be dealing with JSON and not CSV input.  No changes necessary, but if you could think of an easy way to abstract the code so its format (json, csv, tsv, etc) agnostic, that would be really helpful.  This is also something we can iterate on in the future.

tests/algorithm/scripts/expand_test_data.py
```diff
@@ -0,0 +1,243 @@
...
+test_cases = []
+with open(BASE_TEST_FILE, 'r') as fp:
+    test_cases = fp.readlines()
+
+# Now, let's do a bit of formatting and parsing:
+# Each case will be a dict mapping a formatted value to the
+# appropriate header
+headers = [h.strip() for h in test_cases[0].split(",")]
+test_cases = test_cases[1:]
+# Some field values are double-quoted strings which themselves contain commas
```
ericbuckley, line 180: Just so you know, the [csv module has a reader](https://docs.python.org/3/library/csv.html#csv.reader) that does this same logic.

ericbuckley (approved):

tests/algorithm/scripts/expand_test_data.py
bamader: Yeah I suspect we'll need to iterate on this as we get into load testing, when we see what our data shapes and structures look like more precisely, but definitely agree. At its heart this is really an intelligent randomization function that keeps distributions looking reasonable and controls quality, and that's not predicated on data format. We should definitely be able to abstract that randomization to a more general case once we get into it!

Issue #376, opened 2025-05-16
Refactor `expand_test_data.py` to accept PIIRecords


m-goggins:
## Summary
Load testing data will need to be scrambled to account for some matches and some non-matches. The `expand_test_data.py` script does a nice job of scrambling data but it only accepts CSV data that comes from NBS while the load testing data will be `PIIRecods`. We need to update the script to abstract away the CSV-specific parts so that the `apply_field_scrambling` function can accept other formats.

## Acceptance Criteria
- [ ] `expand_test_data.py` is refactored to work on a variety on input data (not just the NBS CSV data)
- [ ]  For the purposes of load testing, it can accept an individual PIIRecord, multiply it approximately 5 times (we need roughly a 5:1 ratio of records to individuals to replicate Montana's data), scramble some of the data, and return it

## Details / Tasks
Note: for the purposes of load testing, it is okay if the individual records are scrambled such that they no longer match the original record.

PR #380, opened 2025-05-16
Add initial locustfile for load testing

### LINKED ISSUE #377: Create locustfile.py for load testing
## Summary
Create a locustfile.py that will run load testing locally. 

## Acceptance Criteria
- [ ] A locustfile.py is added to `tests/load` 
- [ ] The locust file loads seed data
- [ ] On start, the seed data is added to the MPI exactly 1 time by hitting the `/seed` endpoint
- [ ] Random records are selected from the seed data to be randomized 
- [ ] The randomized records run through RL by hitting the `/link` endpoint

m-goggins:
## Description
This PR adds an initial `locustfile` for running load testing initially. It will need to be modified as the load testing plan develops but this gives you an idea of how load testing might work locally. 

**How to run the load test**:
For now (eventually will make this a single script):
1. Spin up the API locally. I used `./scripts/local_server.sh` 
2. Run `locust -f ./tests/load/locustfile.py --headless -u1 -r 1 --run-time 3s --host http://localhost:8000`

**Interpreting the output**:
You will get two sets out output after the load test has completed:
1. # of requests, # fails, and time for each endpoint that is hit
2. Approximate response time percentiles 

We can save the output as a CSV for easier sharing with `--csv example`

**Details about locustfile**:
1. The seed data is loaded: I went with loading a simple file for now, but as we increase the number of records, we may need to read in the file incrementally in case it is too large to open in memory or generate the seed data on the fly.
4. On start, the seed data is added into the MPI: currently this runs once per user (not once per test) and will need to be fixed in future iterations. To avoid seeding multiple times, I have set the number of users to 1. 
5. Randomize the seed data: A random record from the seed data is chosen and run through record linkage by hitting the `/link` endpoint. In future iterations, the record will be duplicated a random number of times and randomized once I have updated `expand_test_data` to work with `PIIRecords` (ticket #376) 

## Related Issues
Closes #377 

## Additional Notes
There are likely lots of iterations we can make on this file as we learn more about how locust works so let me know if you find something neat!


src/recordlinker/assets/test_data.json
ericbuckley, line 1: What do you think about putting this file in the tests/load directory if its just going to be used for testing and not production code?

tests/load/locustfile.py
```diff
@@ -0,0 +1,43 @@
...
+
+class LoadTest(locust.HttpUser):
+    seed_data = None
+    seeded = False
+    seed_lock = threading.Lock()
+
+    def load_seed_data(self):
+        locust_file = pathlib.Path(__file__).resolve()
+        project_root = locust_file.parents[2]  # -> goes up to the project root
+        json_path = project_root / "src" / "recordlinker" / "assets" / "test_data.json"
```
ericbuckley, line 17: If you use [os.path.join](https://docs.python.org/3/library/os.path.html#os.path.join) it should work for both unix and windows.

ericbuckley (approved):

tests/load/locustfile.py
```diff
@@ -0,0 +1,43 @@
...
+
+        with open(json_path, "r") as f:
+            return json.load(f)
+
+    def on_start(self):
+        with self.seed_lock:
+            if not self.__class__.seeded:
+                self.__class__.seed_data = self.load_seed_data()
+                self.client.post("/api/seed", json=self.__class__.seed_data)
```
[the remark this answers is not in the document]
m-goggins: Ah good to know! I'll add a TODO so I don't lose track of this

PR #381, opened 2025-05-19
Fixes dependency tree

johanna-skylight:
## Description
While setting up the dibbs landing page I was able to fix the dependency tree and drop the legacy peer deps flag. This PR sets the correct versions for the dependencies triggering errors during `npm` install and removes the `--legacy-peer-deps` flag from the `npm install` commands. 

## Related Issues
None


ericbuckley (approved):
Thanks for resolving this!

PR #388, opened 2025-05-22
spike on background job

### LINKED ISSUE #382: Experiment running a background job in FastAPI
## Summary
For calibration, we'll need the capacity to run a long job.  Depending on what path we take, this could be minutes or hours, but in either case this is well outside the range of an [acceptable API call](https://odown.com/blog/what-is-a-good-api-response-time/#acceptable-api-response-times).  FastAPI has the ability to run [background jobs](https://fastapi.tiangolo.com/reference/background/), we should experiment to see if this is a viable path forward or if we need a more complex architecture for calibration.

## Details / Tasks
NOTE: The following is a [common API design pattern](https://tyk.io/blog/api-design-guidance-long-running-background-jobs/), but not our only option here.  Feel free to experiment with similar procedures if they can also prove that FastAPI can reliably create, complete and report back a long-running job.

Create a new API endpoint `POST /job/test`, this endpoint optionally accepts 1 parameter, `delay`, which is an integer representing the number of seconds the background job should run for.

The API endpoint should return a 202 status code, with a url indicating where the status of the job can be found.  That url will be in the form of `GET /job/status/<id>`, where `id` is a unique identifier for the job that was just initiated.

Another new API endpoint `GET /job/status/<id>` will need to report back on any valid background jobs created.  It should either report back that the job is "in progress" or "complete".  If the job can't be found, return a 404 status code.

With the above endpoints, we should be able to write a script that initiates a new background job, and then polls it for a status update every X seconds.  Once the polling returns "complete", the script can exit.  The API should be running in a docker container to simulate how this would work in a production environment.

ericbuckley:
**NOTE: DO NOT MERGE**
 
## Description
A POC PR to show how we could initiate a background calibration job and monitor its completion.  

## Related Issues
closes #382 

## Additional Notes
This PR is not intended to be merged. It’s being opened solely for review and discussion, so others can provide feedback on the design and share their thoughts on whether it aligns with our actual calibration procedure.

## Checklist

## Checklist for Reviewers

### DIFF
--- a/scripts/example_linkage_test.sh
+++ b/scripts/example_linkage_test.sh
@@ -261,3 +261,22 @@ PERSON_A_ID=$(echo "$RESPONSE_BODY" | python3 -c 'import sys, json; print(json.l
 echo "ASSIGN RECORD TO CLUSTER..."
 read -r HTTP_STATUS RESPONSE_BODY <<< "$(http_request "${SERVER}/person/${PERSON_A_ID}" "PATCH" "{\"patients\": [\"${PATIENT_ID}\"]}")"
 assert_status_code "$HTTP_STATUS" "200"
+
+echo "CREATE CALIBRATION JOB..."
+read -r HTTP_STATUS RESPONSE_BODY <<< "$(http_request "${SERVER}/calibration/test" "POST" "{\"delay\": 120}")"
+assert_status_code "$HTTP_STATUS" "202"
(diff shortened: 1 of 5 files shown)


ericbuckley:
closing this.  The spike proved that background jobs do work as expected, we can reference this PR and the closed branch in the future if we need to do future work like this.

PR #394, opened 2025-05-30
updates to locustfile

### LINKED ISSUE #393: Update locustfile to run load test

m-goggins:
## Description
This PR updates the locustfile for load testing to:

- Check whether the MPI has been seeded based on user input
- Add a flag for number of records to link to better control the load testing size
- Loop through the original data (1.5 M records), randomly choose a record, and send to the link endpoint. Repeat until the number of records to link has been reached

## Related Issues
Closes #393 

## Additional Notes
If you have better ideas for randomly selecting records from the original data, I'm all ears, because I'm worried it will make seeding artificially slow. Right now, the `original_data` file is too large to load into memory to allow for random indexing so it loops through the records and randomly decides if they should be linked. It seems unlikely that we'd test all 1.5M records so the first records would be more heavily tested. This probably doesn't matter for load testing but I wanted to point it out. 

The other option I considered (but ruled out) was to create the list of records that would be linked during the seeding process, i.e., when looping through the `clusters_iter` in the `on_start` section, you could also randomize whether a cluster gets added to a `record_to_link` list. However, we'd likely run into the same issue with the list being too large to hold in memory. We could write the list to file and then read from that in the `link` task, but it would only remove line 85 `if random.random() < 0.5`. 

Curious to hear other ideas!

## Checklist

## Checklist for Reviewers

### DIFF
--- a/tests/load/locustfile.py
+++ b/tests/load/locustfile.py
@@ -4,42 +4,91 @@ import pathlib
 import random
 import threading
 
+import ijson
 import locust
+
+
+def str_to_bool(value):
+    """
+    Helper function to convert a string representation of truth to a boolean value in
+    locust CLI.
+    """
+    if isinstance(value, bool):
+        return value
+    if value.lower() in {"true", "1", "yes"}:
+        return True
+    elif value.lower() in {"false", "0", "no"}:
+        return False
+    else:
+        raise ValueError(f"Invalid boolean value: {value}")
+
+
+@locust.events.init_command_line_parser.add_listener
+def _(parser):
+    parser.add_argument(
+        "--records-to-link", type=int, default=100, help="Number of records to link in the test run"
+    )
+    parser.add_argument(
+        "--seeded",
+        type=str_to_bool,
+        default=False,
+        help="Indicates if the database has been seeded with initial data",
+    )
 
 
 class LoadTest(locust.HttpUser):
     seed_data = None
-    seeded = False
     seed_lock = threading.Lock()
 
-    def load_seed_data(self):
-        locust_file = pathlib.Path(__file__).resolve()
-        project_root = locust_file.parents[2]
-        json_path = os.path.join(project_root, "tests/load/assets/test_data.json")
-
-        with open(json_path, "r") as f:
-            return json.load(f)
-
     def on_start(self):
+        # Check if the database has been seeded before starting the test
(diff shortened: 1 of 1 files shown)


tests/load/locustfile.py
```diff
@@ -4,42 +4,91 @@
...
+        # Pick a random record from the seed data
+        locust_file = pathlib.Path(__file__).resolve()
+        project_root = locust_file.parents[2]
+        original_data_path = os.path.join(project_root, "tests/load/assets/original_data.json")
+
+        with open(original_data_path, "rb") as input_file:
+            clusters_iter = ijson.items(input_file, "clusters.item", use_float=True)
+            counter = 1
+            for cluster in clusters_iter:
+                if random.random() < 0.5:  # randomly decide whether to link or not
```
ericbuckley, line 85: With this implementation, we'll send about 50% of the data to be linked.  What do you think about making this a variable?

ericbuckley (commented):
Looking good, you can just pin mypy to 1.15.0 in pyproject.toml to get around the type checking issue.

pyproject.toml
```diff
@@ -7,16 +7,16 @@ name = "RecordLinker"
 dynamic = ["version"]
 description = "An API for linking health care records using Patient PII"
 authors = [
-    {name = "Brandon Mader", email = "brandon@skylight.digital"},
```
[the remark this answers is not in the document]
m-goggins: All formatting with the exception of pinning mypy now

tests/load/locustfile.py
```diff
@@ -4,42 +4,98 @@
...
+                            chunk = []
 
-    def pick_random_record(self):
-        # Pick a random record from the seed data
-        rando = random.choice(self.__class__.seed_data["clusters"])
-        return rando
+                    if chunk:
+                        seed_data = json.dumps({"clusters": chunk}, indent=2)
+                        self.client.post("/api/seed", seed_data)
+                self.__class__.seeded = True
```
ericbuckley, line 73: I'm new to locust, but does this value need to be persisted back to `self.environment`?  I'm not seeing how `self.environment.parsed_options.seeded` would return True on subsequent runs.

ericbuckley: wfm

ericbuckley (approved):

PR #441, opened 2025-06-18
Fixing record parsing issues with empty strings and null values 

ericbuckley:
## Description
Fixing cases in which a null value or empty string in a record payload can cause a 500.  

## Additional Notes

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/schemas/pii.py
+++ b/src/recordlinker/schemas/pii.py
@@ -282,13 +282,14 @@ class Address(StrippedBaseModel):
         normalized: list[str] = []
         if value:
             for line in value:
-                parts = line.strip().split(" ")
-                # remove all non-alphanumeric characters and convert to uppercase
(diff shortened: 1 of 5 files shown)


src/recordlinker/schemas/pii.py
```diff
@@ -347,6 +348,22 @@ def validate_and_normalize_telecom(self) -> typing.Self:
 
         return self
 
+    def normalized_value(self) -> str:
```
ericbuckley, line 351: @m-goggins I created this function to DRY up some code in the field_iter, basically the logic we were using for parsing TELECOM, PHONE and EMAIL.  While doing this, I was wondering, should we?  That is, should we just be normalizing the phone number at ingestion and storing the national number at that point, so we don't have to run `phonenumbers.parse(self.value).national_number` in the field_iter?  What do you think?

m-goggins (approved):

PR #449, opened 2025-06-25
Add RMS boundary

### LINKED ISSUE #446: Calculate suggested RMS boundaries

bamader:
## Description
This PR adds functionality for scoring each of the class-partitioned pairs of patient records from the DB, and then using those scores to estimate reasonable bounds for the possible match window (MMT and CMT). Numerous helper functions got added to tuning to allow us to perform a modified version of RL's evaluation step (no blocking needed because the records are already in pairs) with us providing manual overrides to a number of parameters (namely log-odds).

## Related Issues
Closes #446 

## Additional Notes
As usual, I set up the local server to add this processing step to hitting the dummy tuning endpoint, and end to end, the entire computation was ~90 seconds. So that means DB sampling and log-odds calculating is 60-70s and scoring and sorting the class lists takes about 20s.

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/tuning/prob_calc.py
+++ b/src/recordlinker/tuning/prob_calc.py
@@ -2,6 +2,8 @@ import math
 import typing
 
 from recordlinker.linking.matchers import compare_probabilistic_exact_match
+from recordlinker.linking.skip_values import remove_skip_values
+from recordlinker.schemas import algorithm as ag
 from recordlinker.schemas.pii import Feature
 from recordlinker.schemas.pii import FeatureAttribute
 from recordlinker.schemas.pii import PIIRecord
@@ -23,7 +25,7 @@ def calculate_class_probs(
     This function is used to calculate both the m- and u-probabilities,
     depending on the pair sample it is given.
 
-    :param true_match_pairs: A sequence of tuples containing pairs 
+    :param sampled_pairs: A sequence of tuples containing pairs 
       of patient data dictionaries.
     :returns: A dictionary mapping Feature names to their class
       probabilities.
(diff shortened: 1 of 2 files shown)


bamader:
@ericbuckley Oh wow, you're completely right. I had a complete brain fart building out the helper functions to sort and score the class partitions because they derive from similar code in `link.py` and there we take the max across all passes. But not here! We absolutely want one possible match window per pass. The control logic should be:

* [...]
* retrieve default algorithm
* for each pass in the default algorithm:
- calculate and sort tuning scores
- estimate RMS bounds using computed scores

Thanks for catching that! I'll get it fixed up.

src/recordlinker/tuning/prob_calc.py
```diff
@@ -68,4 +70,219 @@ def calculate_log_odds(
...
+                break
+        for t in true_match_scores:
+            if t > non_match_scores[-1]:
+                cmt = t
+                break
+
+        # To account for unseen data, buffer each threshold by pushing it 
+        # towards its respective distribution's extreme
+        if mmt is not None:
+            mmt = max([0, mmt - 0.025])
```
ericbuckley, line 162: Why lower the mmt by 0.025 and raise the cmt by that same amount?

ericbuckley (approved):
looks good!

src/recordlinker/tuning/prob_calc.py
ericbuckley: 1) Is it worth logging any of these intermediate values so we can see what it looks like with production data?
2) Is the shift a function of the sample size?  For instance, if a jurisdiction configured the app with a small amount of true matches and non matches, would we want to increase this shift?

bamader: 1. We can but I don't think there's a ton of value in doing so. The heuristics are very easy to reverse if needed, and since we're not doing anything graphically (plotting the curves) the numbers will be less helpful to visually inspect because they won't be in the context of distributional score curves (e.g. we can't see how much the tails overlap, how "tall" the overlap is from a histogram frequency perspective, how wide the distributional overlap is overall, etc.). If we get to a place where we actually do plot the curves I think looking at intermediate values is _much_ more valuable though.
2. I initially did have it be a function of sample size that shifted outward proportional to the extent of the overlap, but I was looking at the code and decided it was way too unnecessarily convoluted to be of any use to a jurisdiction. It was a whole lot of numbers and proportional scaling based on differential ranges and I was like no one _actually_ needs this to be effective, just go simpler. It was a real Occam's Razor situation (I also did a bit more mathing to check myself and it was too hard to control the shift amount because it's not just a function of the sample size, but also the degree of overlap, so it was just a whole thing to compute a value difference that frankly won't change many scores if it's 0.005 vs 0.022).

PR #458, opened 2025-07-08
Fix locust testing file

m-goggins:
## Description
Fixes a couple of errors I encountered while doing the 20M patient record load testing that I forgot I had stashed.

## Related Issues
n/a

## Additional Notes
I added back the `--seeded` option and made the `--record-data` option optional for the case where locust is used just for seeding.

### DIFF
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -57,7 +57,7 @@ jobs:
           elif [[ "${{ matrix.database }}" == "mysql" ]]; then
             export DB_URI="mysql+pymysql://root:pw@localhost:3306/mysql"
           elif [[ "${{ matrix.database }}" == "sqlserver" ]]; then
-            export DB_URI="mssql+pyodbc://sa:YourStrong!Passw0rd@localhost:1433/master?driver=ODBC+Driver+18+for+SQL+Server&TrustServerCertificate=yes"
+            export DB_URI="mssql+pyodbc://sa:YourStrong!Passw0rd@localhost:1433/master?driver=FreeTDS&TrustServerCertificate=yes&TDS_Version=8.0"
           fi
 
           API_ROOT="/api"
--- a/.github/workflows/check_smoke_tests.yml
+++ b/.github/workflows/check_smoke_tests.yml
@@ -57,7 +57,7 @@ jobs:
           elif [[ "${{ matrix.database }}" == "mysql" ]]; then
             export DB_URI="mysql+pymysql://root:pw@localhost:3306/mysql"
           elif [[ "${{ matrix.database }}" == "sqlserver" ]]; then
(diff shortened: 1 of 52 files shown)


.github/workflows/check_smoke_tests.yml
```diff
@@ -57,7 +57,8 @@ jobs:
           elif [[ "${{ matrix.database }}" == "mysql" ]]; then
             export DB_URI="mysql+pymysql://root:pw@localhost:3306/mysql"
           elif [[ "${{ matrix.database }}" == "sqlserver" ]]; then
-            export DB_URI="mssql+pyodbc://sa:YourStrong!Passw0rd@localhost:1433/master?driver=ODBC+Driver+18+for+SQL+Server&TrustServerCertificate=yes"
+            export DB_URI="mssql+pyodbc://sa:YourStrong!Passw0rd@localhost:1433/master?driver=FreeTDS&TrustServerCertificate=yes"
```
ericbuckley, line 60: I was thinking the driver should continue to match what we have documented in docs/site/app-configuration.md.  That way we know what we are telling people to use still works.  Thoughts?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.12-slim
+FROM python:3.13-alpine
```
ericbuckley, line 1: I'm a bit confused with this diff.  In main, we're [currently using python:3.13-alpine](https://github.com/CDCgov/RecordLinker/blob/main/Dockerfile#L1).

pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
     "psycopg2-binary", # PostgreSQL
     "PyMySQL",         # MySQL & MariaDB
     "cryptography",    # MySQL & MariaDB dependency
-    "pyodbc",          # Microsoft SQL Server
+    "pyodbc>=4.0.35",  # SQL Server
```
ericbuckley, line 36: 5.2.0 is the current version.  Is there a situation in which you think a pre 4.0.35 build could get installed?

tests/load/locustfile.py
```diff
@@ -81,7 +87,7 @@ class LoadTest(locust.HttpUser):
     def on_start(self):
         options = self.environment.parsed_options
         if options.seed_data:
-            if os.path.exists(options.seed_data):
+            if not os.path.exists(options.seed_data):
```
ericbuckley, line 90: good catch

.github/workflows/check_smoke_tests.yml
m-goggins: You're totally right

Dockerfile
m-goggins: I don't think all the changes from main had been pulled. This has been resolved with the latest pull/push

PR #469, opened 2025-07-16
proc calc with iterators

### LINKED ISSUE #456: Reduce tuning memory used
## Summary
During a standard run (default tuning params) with a database of 500k person clusters and around 2 million patient records, the system uses up to 1.25GB of RAM.  We should explore ways to reduce this footprint so that NBS can run this feature in lower environments with less resources.

ericbuckley:
## Description
Optimizing memory usage by ~70% in tuning calculations using iterators.

## Related Issues
closes #456

## Additional Notes
- I couple of dataclasses were added to schemas/tuning to help with passing data between prob_calc and base.  TuningPair and TuningProbabilities.  The former is particularly helpful with a shift to iterators, as it's no longer possible to return values from the mpi_service queries.
- AlgorithmPass.resolved_label was added to help with type checks, as that guarantees a string will be returned (AlgorithmPass.label, could be null)
- Feature was configured to be "frozen" (meaning once a Feature object is instantiated, it can't be changed).  This is useful as we can now use them as keys in our log-odds dicts, rather than the feature string values as the keys.  This reduced the number of time we need to convert back and forth between strings and Features.
- tuning/base::tune was split to use 2 sub-functions, run_log_odds and run_rms.  With an iterators version, we need to gather true-match and non-match pairs from the database twice.  Separating the calculations into two functions makes it clearer as to where the pairs are being used.
- The looping constructs in calculate_and_sort_tuning_scores we essentially reversed.  Previously we looped over all the algorithm passes, then we looped over the pairs.  Now that we have iterators, that doesn't work, as we can only access the pair once.  To accomplish this, we now loop over the pairs, then loop over the algorithm passes.  _score_pairs_in_class was renamed to _score_records_in_pair as now it just performs the value on 1 pair over N passes.

## Performance Notes
Results from running tuning job tests with a 500k cluster using the default params.

### main stats
runtime: 42s
max memory: 665MB

### PR-469 stats
runtime: 44s
max memory: 204MB

## Checklist

## Checklist for Reviewers

### DIFF
--- a/src/recordlinker/schemas/tuning.py
+++ b/src/recordlinker/schemas/tuning.py
@@ -119,9 +119,10 @@ class TuningPair:
 @dataclasses.dataclass
 class TuningProbabilities:
     """
-    A dict of feature specific probabilities that two records are likely to belong to
-    the same cluster given a collection of TuningPairs. Also included are the number,
-    count, of TuningPairs analyzed and an optional parameter for sample size used.
+    A dictionary of class-conditional likelihoods separated by feature. For a given
+    feature F, this holds the probability that a pair of patient records in the same
+    tuning class (i.e. a true-match pair or a non-match pair) will have the same field
+    value in F.
     """
     probs: dict[Feature, float]  # the feature specific probabilities
     count: int  # the number of pairs analyzed
--- a/src/recordlinker/schemas/tuning.py
+++ b/src/recordlinker/schemas/tuning.py
@@ -5,6 +5,7 @@ recordlinker.schemas.tuning
 This module contains the schema definitions for the tuning jobs.
 """
 
+import dataclasses
 import datetime
 import typing
 import uuid
@@ -17,6 +18,8 @@ from recordlinker import models
 from recordlinker.utils.datetime import now_utc_no_ms
 
 from .algorithm import LogOdd
+from .pii import Feature
+from .pii import PIIRecord
 
 
 class TuningParams(pydantic.BaseModel):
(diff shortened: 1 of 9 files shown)


src/recordlinker/schemas/tuning.py
```diff
@@ -90,3 +93,36 @@ def from_tuning_job(cls, job: TuningJob, request: fastapi.Request) -> typing.Sel
...
+            sample_used=sample_used,
+        )
+
+
+@dataclasses.dataclass
+class TuningProbabilities:
+    """
+    A dict of feature specific probabilities that two records are likely to belong to
+    the same cluster given a collection of TuningPairs. Also included are the number,
+    count, of TuningPairs analyzed and an optional parameter for sample size used.
```
bamader, line 124: This docstring isn't quite accurate (it confused me on first read because I was trying to figure out why we had the probability that two records belong to the same class). When I got to prob_calc, I realized what this data class is for though, so there's an easy fix if we just borrow the language used in those docstrings (helps with both consistency and correctness). I would do something like:

```
A dictionary of class-conditional likelihoods separated by feature. For a given feature F, this holds the probability that a pair of patient records in the same tuning class (i.e. a true-match pair or a non-match pair) will have the same field value in F.
```

src/recordlinker/schemas/tuning.py
```diff
@@ -90,3 +93,36 @@ def from_tuning_job(cls, job: TuningJob, request: fastapi.Request) -> typing.Sel
...
+    """
+    probs: dict[Feature, float]  # the feature specific probabilities
+    count: int  # the number of pairs analyzed
+    sample_used: typing.Optional[int] = None  # the number of records sampled from to produce the probabilities
```
bamader, line 128: What's the thinking on including these as variables in the data class rather than just tracking them in the function in which they're used? For example, in `calculate_class_probs` in prob_calc, we could start a variable `count = 0` outside the for loop over the iterator, increase it by 1 each iteration, and then divide the dictionary values here by `count` at the bottom of the function without actually storing count in the loop. Any reason to preserve it?

src/recordlinker/tuning/prob_calc.py
```diff
@@ -98,36 +97,40 @@ def calculate_and_sort_tuning_scores(
...
+    }
+
+    # Both true-match and non-match pairs are iterables and can only be accessed once,
+    # we need to use the pairs to calculate values for all passes in the Algorithm, thus we
+    # need to iterate over the pairs first, then the passes
+    for pair in true_match_pairs:
+        for key, score in _score_records_in_pair(pair, log_odds, max_points, algorithm).items():
+            if score > 0.0:
+                # ignore true-matches with no agreement
+                sorted_scores[key][0].append(score)
```
bamader, line 115: Nit: Just thinking about possible downstream modifications, we may want to leave this in `estimate_rms_bounds` instead of pre-filtering here. We know there's the possibility that NBS adds some more analysis around tuning at some point (based on Julie's asks to separate log-odds calculation from RMS calculation), and if as part of that they want to look at the scores for the true-match class it would probably be useful to have the 0's included so they could analyze the distribution. Taking the 0s out only really matters for the RMS bounds so my vote would be to leave it there. Non-blocker though!

bamader (approved):
Looks like a good clean refactor! Just a couple of minor comments around docstrings and an organizational questions, but nothing that I think is blocking.

src/recordlinker/tuning/prob_calc.py
ericbuckley: ok.  Just thought it might be a little optimization to remove them early if we just discard later.  I can revert though.

